PR #4, opened 2018-06-19
Improved bootstrap config management

toonsevrin:
Here you go, the first PR! Already?!

This does however require a quick peer review and test run to verify it works (I didn't do this...).

Perhaps gradle needs to be configured for the resource folder but I don't think so. I also think the files should be accessible from the tests from default but this should of course be tested.


talekhinezh:
@toonsevrin yoyo! Nice to see first PR already! Some thoughts:
* Agree that json file as config makes better sense
* Putting it in `src/resources/main` makes sense. Only issue is this lib is being used for Android as well which I believe can't look into those folders. Do you know of a good workaround this (I'm an android noob).
* `IllegalStateException` is probably the wrong exception in this case. Probably `IOException` makes more sense.
* I don't think files should be accessible from tests as unit tests should be as IO independent as possible.

toonsevrin:
It's actually very likely that the android gradle plugin actually does that resource management for us. So that part would already be taken care of.

talekhinezh:
@MarcRubio Can you check out this android gradle plugin @toonsevrin is suggesting to make sure it works for you?

radixdlt-java/src/main/java/com/radixdlt/client/core/address/RadixUniverseConfigs.java
```diff
@@ -1,30 +1,27 @@
...
 	public static final RadixUniverseConfig getHighgarden() {
-		return RadixUniverseConfig.fromDson(highgardenUniverseDson);
+		return RadixUniverseConfig.fromFile(getConfigFile("highgarden.json"));
 	}
 
 	public static final RadixUniverseConfig getAlphanet() {
-		return RadixUniverseConfig.fromDson(alphanetUniverseDson);
+		return RadixUniverseConfig.fromFile(getConfigFile("alphanet.json"));
+	}
+	private static File getConfigFile(String name){
```
MarcRubio: Using getResourceAsStream deals with the Android issue

suggestion:
```
private static InputStream getConfigFileStream(String name) {
		String source = File.separator + "configs" + File.separator + "bootstrap" + File.separator + name;
		return RadixUniverseConfig.class.getResourceAsStream(source);
	}
```

radixdlt-java/src/main/java/com/radixdlt/client/core/address/RadixUniverseConfigs.java
```diff
@@ -1,30 +1,27 @@
...
+	private RadixUniverseConfigs() {}
 
 	public static final RadixUniverseConfig getWinterfell() {
-		return RadixUniverseConfig.fromDson(testUniverseDson);
+		return RadixUniverseConfig.fromFile(getConfigFile("testuniverse.json"));
```
MarcRubio: get the stream instead of file
`return RadixUniverseConfig.fromInputStream(getConfigFileStream("testuniverse.json"));`

radixdlt-java/src/main/java/com/radixdlt/client/core/address/RadixUniverseConfig.java
```diff
@@ -25,7 +31,17 @@ public static RadixUniverseConfig fromDson(String universeDson) {
 		RadixUniverseConfig universe = RadixJson.getGson().fromJson(jsonElement, RadixUniverseConfig.class);
 		return universe;
 	}
+	
+	public static RadixUniverseConfig fromFile(File file){
```
MarcRubio: Instead of dealing with the file use the stream.

```
public static RadixUniverseConfig fromInputStream(InputStream inputStream) {
		try {
			return RadixJson.getGson().fromJson(new InputStreamReader(inputStream), RadixUniverseConfig.class);
		} catch (Exception e) {
			throw new IllegalStateException("Failed to read stream from bootstrap file");
		}
	}
```

toonsevrin:
@MarcRubio I'm travelling for the upcoming 6 weeks so won't be able to check your patches. If they work, feel free to implement them.

Issue #31, opened 2018-06-30
Add JSON support for retrieving universe information


murraco:

talekhinezh:
Fixed

PR #51, opened 2018-07-03
Add JSON support for retrieving universe information

murraco:
#31 - The information of the universe can be retrieved in JSON format

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
@@ -1,6 +1,5 @@
 package com.radixdlt.client.core;
 
-import com.google.gson.JsonObject;
 import com.radixdlt.client.core.address.RadixAddress;
 import com.radixdlt.client.core.address.RadixUniverseConfig;
 import com.radixdlt.client.core.address.RadixUniverseConfigs;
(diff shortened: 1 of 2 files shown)


radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
```diff
@@ -171,4 +172,8 @@ public void disconnect() {
 		ledger.close();
 		network.close();
 	}
+
+	public JsonObject toJson() {
```
talekhinezh: `RadixUniverse`, as defined in the lib as ledger+network, should not be serializable. Instead, expose method to retrieve config object.

radixdlt-java/src/main/java/com/radixdlt/client/core/address/RadixUniverseConfig.java
```diff
@@ -66,6 +67,20 @@ public RadixAddress getSystemAddress() {
 		return genesis;
 	}
 
+	public JsonObject toJson() {
+		JsonObject universe = new JsonObject();
```
talekhinezh: Can just do this with `RadixJson.getGson().toJsonTree(this)` no?

murraco: Yes but if I do it that way it returns `JsonElement` instead of a `JsonObject`

radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
murraco: I updated this

PR #58, opened 2018-07-14
Decimal format for assets that have a fixed number of decimals

nadam:
Writing test rads in decimal format in InsufficientFundsException, while keeping fraction format for assets that don't divide to a fixed number of decimals.

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/assets/Asset.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/assets/Asset.java
@@ -15,7 +15,6 @@ public class Asset {
 	private final String iso;
 	private final int subUnits;
 	private final EUID id;
-	private final boolean powerOfTen;
 
 	public Asset(String iso, int subUnits, EUID id) {
 		Objects.requireNonNull(iso);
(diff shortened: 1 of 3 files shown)


radixdlt-java/src/main/java/com/radixdlt/client/assets/Asset.java
```diff
@@ -52,4 +61,11 @@ public boolean equals(Object o) {
 	public int hashCode() {
 		return iso.hashCode();
 	}
+
+	private boolean isPowerOfTen(int value) {
```
talekhinezh: I think this method would make more sense to be placed in AssetAmount.java as isPowerOfTen() is too generic of a method for an Asset. An *Asset* cannot be a power of ten but it's subunits property can.

nadam:
Ok, I had it in AssetAmount.java first, but moved it to Asset.java since that makes it only need to be calculated once. A bit premature optimization I guess :)

I'll move it back again.

talekhinezh:
@nadam Looks good to me. Looks like there is just an issue with sonarcloud for some reason though the error is unclear to me. @blenessy can you take a look at this?

blenessy:
Not sure what happened here. I tried to re-run the failed build but I think whatever broke got cached by travis because every time I re-run I get the same problem. Re-running the checks on master work as expected. 

Conclusively, I'm merging this because the test pass and that is most important.

PR #61, opened 2018-07-19
Jwt service

talekhinezh:
Updates to the library to enable a jwt service to be written on top.


talekhinezh:
@blenessy something happened with this build as well. Looks like a different error as well. Going to merge this though as it passes tests locally and this is needed for the Hackathon.

PR #62, opened 2018-07-23
Support for alternative key storage using Reader and Writer

nadam:
The implementation assumed that the keys should be stored in a file. With this change they can be stored elsewhere, for instance in a database.

Also fixed some exception declarations.


radixdlt-java/src/main/java/com/radixdlt/client/core/identity/PrivateKeyEncrypter.java
```diff
@@ -96,7 +100,8 @@ public static void createEncryptedPrivateKeyFile(String password, String filePat
         byte[] computedMac = generateMac(derivedKey.getEncoded(), cipherText);
 
         if (!Arrays.equals(computedMac, ByteString.decodeHex(mac).toByteArray())) {
-            throw new Exception("MAC mismatch");
+            // TODO: Change to MacMismatchException and make that exception extend GeneralSecurityException?
```
MarcRubio: The crypto package already has a MacMismatchException which you can use though it extends the Exception class but you are right, it would make sense to throw this kind of exception so you can go ahead and make this change IMO and remove the TODO

nadam:
So I did a rebase and added the same changes again in a single commit. I hope it's still ok.

PR #63, opened 2018-07-27
Add onErrorComplete in connect method

MarcRubio:
This takes care of a OnErrorNotImplemented which can happen when attempting to connect.


radixdlt-java/src/main/java/com/radixdlt/client/core/network/WebSocketClient.java
```diff
@@ -100,7 +100,8 @@ public Completable connect() {
 			})
 			.filter(status -> status.equals(RadixClientStatus.OPEN))
 			.firstOrError()
-			.ignoreElement();
+			.ignoreElement()
+			.onErrorComplete();
```
talekhinezh: I don't think we want this here as we don't want this flow to complete if an error has occurred. The error should be handled downstream, no?

MarcRubio:
Close since refactor will likely handle things differently.

PR #73, opened 2018-08-24
Metadata

talekhinezh:
Add metadata property into Data Particle.

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/application/translate/DataStoreTranslator.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/application/translate/DataStoreTranslator.java
@@ -40,7 +40,7 @@ public class DataStoreTranslator {
 		Map<String, Object> metaData = new HashMap<>();
 		metaData.put("timestamp", atom.getTimestamp());
 		metaData.put("signatures", atom.getSignatures());
-		metaData.put("application", atom.getDataParticle().getApplication());
+		metaData.put("application", atom.getDataParticle().getMetaData("application"));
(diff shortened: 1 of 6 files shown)


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/DataStoreTranslator.java
```diff
@@ -40,7 +40,7 @@ public Completable translate(DataStore dataStore, AtomBuilder atomBuilder) {
 		Map<String, Object> metaData = new HashMap<>();
 		metaData.put("timestamp", atom.getTimestamp());
 		metaData.put("signatures", atom.getSignatures());
-		metaData.put("application", atom.getDataParticle().getApplication());
+		metaData.put("application", atom.getDataParticle().getMetaData("application"));
```
mpsandiford: I'm not sure this is ideal - it will create a `[key, null]` pair in the map, which is probably not desirable for something that is missing, as calls to `containsKey(...)` will return true, but `get(...)` will return `null`.

Could use something like `metaData.compute("application", (k, v) -> atom.getDataParticle().getMetaData("application"))` which will not add a key with a `null` value.

talekhinezh: good point, that works.

talekhinezh: though this logic should be refactored down the line, so that inserting a null value isn't exposed.

PR #76, opened 2018-09-01
Unique property

talekhinezh:
Add application layer unique property which allows transactions to have a user specified unique id which no other transaction can have.


talekhinezh:
@MarcRubio please review.

PR #87, opened 2018-09-21
Simplify API naming scheme of native wallet

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/dapps/wallet/RadixWallet.java
```diff
@@ -107,7 +107,7 @@ public SendResult send(
 		@NonNull RadixAddress toAddress
 	) {
 		Objects.requireNonNull(toAddress, "toAddress must be non-null");
-		return this.send(amount, toAddress, null);
+		return this.send(amount, null, toAddress);
```
MarcRubio: Changed the ordering of params for the send method therefore sendWhenAvailable should also be changed for consistency.

talekhinezh: Good point, fixed.

PR #99, opened 2018-10-12
Fixed Supply Tokens

talekhinezh:
Add support for Fixed Supply Tokens along with a lot of refactoring.


examples/src/main/java/com/radixdlt/client/examples/RadixWalletExample.java
```diff
@@ -38,28 +37,33 @@ public static void main(String[] args) throws Exception {
 			.getStatusUpdates()
 			.subscribe(System.out::println);
 
-
 		RadixApplicationAPI api = RadixApplicationAPI.create(radixIdentity);
```
flotothemoon: <nitpicking> Should we capitalize it RadixApplicationApi instead for consistency? </nitpicking>

flotothemoon:
I would suggest making the actions stand on their own - verbs in class names seem a bit awkward to me. So instead of TransferTokens we could call it TransferTokensAction etc

examples/src/main/java/com/radixdlt/client/examples/RadixWalletExample.java
talekhinezh: Not a fan of changing signatures at the moment as will probably change it down the line again as architecture gets more defined to something like `RadixApplication`

talekhinezh:
@flotothemoon agreed.

PR #106, opened 2018-10-23
The DSON/JSON "serializer" field is now included when calculating the hash.

mpsandiford:
Refactored "version" and "serializer" required fields to be in a common base `SerializableObject` class.

Adjusted outputs for "serializer" field to always be included.

### DIFF
--- a/radixdlt-java/src/main/java/org/radix/serialization2/client/SerializableObject.java
+++ b/radixdlt-java/src/main/java/org/radix/serialization2/client/SerializableObject.java
@@ -1,0 +1,23 @@
+package org.radix.serialization2.client;
+
+import org.radix.serialization2.DsonOutput;
+import org.radix.serialization2.DsonOutput.Output;
+import org.radix.serialization2.SerializerDummy;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Class with serializable fields for version and serializer
+ * that can be inherited to easily implement those fields in
+ * a standard way for DSON and JSON output.
+ */
+public class SerializableObject {
+	@JsonProperty("version")
+	@DsonOutput(Output.ALL)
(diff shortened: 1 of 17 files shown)


radixdlt-java/src/main/java/org/radix/serialization2/client/SerializableObject.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Class with serializable fields for version and serializer
+ * that can be inherited to easily implement those fields in
+ * a standard way for DSON and JSON output.
+ */
+public class SerializableObject {
+	@JsonProperty("version")
+	@DsonOutput(Output.ALL)
+	private short version = 100;
```
flotothemoon: Where should inheriting classes set their version?

mpsandiford: I decided to go for a minimal approach.  I think simply providing classes with the ability to specify their version number isn't sufficient, or really even necessary for an effective versioning scheme.  This is really a placeholder for a improved approach in my opinion.

In practice, if classes *really* want to override this, they can provide their own `JsonProperty` for "version".

talekhinezh:
I am a little skeptical about the `SerializableObject` approach. Is this a temporary solution to get it working or a more long term solution?

Thoughts on it:
* Could we get into trouble if we forget to extend `SerializableObject` on a new class and the serializer does not throw an exception?
* Because one of DSON's defining features (compared to say JSON) is we can use DSON as input for a content-based hash, shouldn't the serializer or object type be a first class citizen?

mpsandiford:
After some discussion with @talekhinezh , we decided to implement a check to ensure that all classes with a SerializerId included a "serializer" field in their hierarchy.

PR #116, opened 2018-11-22
Refactor application level modules to utilize an ActionStore

talekhinezh:
One step towards better modularization


radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -255,21 +264,12 @@ public RadixAddress getMyAddress() {
 		return universe.getAddressFrom(identity.getPublicKey());
 	}
 
-	public Observable<Data> getData(RadixAddress address) {
```
flotothemoon: Should we perhaps change "Data" to "Message"? Similarly, "UnencryptedData" to "UnencryptedMessage" etc.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/AtomToActionsMapper.java
```diff
@@ -0,0 +1,9 @@
+package com.radixdlt.client.application.translate;
+
+import com.radixdlt.client.application.identity.RadixIdentity;
+import com.radixdlt.client.core.atoms.Atom;
+import io.reactivex.Observable;
+
+public interface AtomToActionsMapper<T> {
```
flotothemoon: The name would suggest (to me anyway) that the implementations return an `Action`. But for example `AtomToTokenTransferActionsMapper` returns `TokenTransfer` rather than a `TransferTokensAction`. Similarly `AtomToMessageActionsMapper` returns `UnencryptedData` instead of `SendMessageAction`. So at the least, we might want to consider renaming it.

flotothemoon: Also, should we have a common base class / interface / something to all `Action`s?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: Good idea

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/AtomToTokenTransferActionsMapper.java
```diff
@@ -0,0 +1,127 @@
...
+	private final RadixUniverse universe;
+
+	public AtomToTokenTransferActionsMapper(RadixUniverse universe) {
+		this.universe = universe;
+	}
+
+	public Observable<TokenTransfer> map(Atom atom, RadixIdentity identity) {
+		return Observable.fromIterable(atom.tokenSummary().entrySet())
+			.filter(e -> !e.getKey().equals(universe.getPOWToken()))
+			.flatMapSingle(e -> {
```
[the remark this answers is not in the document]
talekhinezh: Not too concerned about it now as that method obviously needs more refactoring so less value in taking time to figure out components needed for a separate method.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/AtomToActionsMapper.java
flotothemoon: I think even having it just be a marker interface has merit. We should add it.

PR #118, opened 2018-11-27
Add means of deterministic key pair generation

laszlourszuly:
This is a somewhat naive implementation as it doesn't take into
consideration the differences between the "secp256k1" algorithm - used
to generate a regular key pair, and the "sha-256" algorithm - used to
generate the seeded key pair.

There is a theoretic concern that the "sha-256" algorithm doesn't have
the same spread in the full key space as the "secp256k1" has. The latter
is designed to avoid hash collisions as far as possible, the former is
designed to produce deterministic hashes from provided seeds.

Furthermore, the way the seeded key pair is intended to be used, the key
pairs it will produce are even more focused to a certain key sub-space
(the one targeting the human readable seed spectra). This may cause an
elevated risk of hash collisions.

We should investigate how big of an issue this is, if at all, and
address it accordingly.


laszlourszuly:
@talekhinezh: I'm not too comfortable with having the `fromSeed(...)` method directly in the `ECKeyPair` class, basically while it won't derive a key pair from an elliptic curve (I think the "EC" part in `ECKeyPair` is derived from "Elliptic Curve"), but rather from a SHA-256 hash. The question then is where to put this convenience abstraction.

Furthermore I have a more cryptographical concern as well (see commit message). Please prove me wrong :)

talekhinezh:
@laszlourszuly Yes, this seems wrong to me and incomplete. To me, the crux of the task is to randomly generate a mnemonic which then maps to seed. A single pass via "sha-256" definitely seems lacking. See: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/ECKeyPairGenerator.java
```diff
@@ -92,4 +93,43 @@ public ECKeyPair generateKeyPair(int numBits) {
...
 
+    /**
+     * Generates a new, deterministic {@code ECKeyPair} instance based on the
+     * provided seed.
+     *
+     * @param seed The seed to use when deriving the key pair instance.
+     * @return A key pair that corresponds to the provided seed.
+     * @throws IllegalArgumentException if the seed is empty or a null pointer.
+     */
+    public ECKeyPair generateKeyPairFromSeed(byte[] seed) {
```
laszlourszuly: It feels a bit awkward to put this method in this class as it won't source an `ECKeyPair` from an elliptic curve, but rather from a `sha256` hash. The question can be taken further by asking if an `ECKeyPair` sourced from a hash really is an `ECKeyPair` or rather something else ("`HashKeyPair`"?)

laszlourszuly:
@talekhinezh Could you please have a second look at this? Do you agree with the suggested design changes (basically moving `fromSeed` from the `ECKeyPair` to `ECKeyPairGenerator`)?

Let's leave the seed hash construction to a different discussion.

talekhinezh:
I'm still not comfortable with that move as like you mention it doesn't have anything to do with generating an ECKeyPair. Even if we are to move forward with using a hash to determine a private key (I'm still not convinced), it should be done in a layer above all the ECKeyPair* classes which is implementation specific, perhaps a new class like `RadixECKeyPairs`.

laszlourszuly:
@talekhinezh I've isolated this unholy spawn of satan a bit more, as suggested. I fully agree that this topic isn't exhausted yet. I suggest we agree on an interim solution that satisfies the requirements of the user story as is now, and open another story to look into an architecturally more sustainable solution. What do you think? And would the currently suggested isolation in this pull request work as an interim solution you think?

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/RadixECKeyPairs.java
```diff
@@ -0,0 +1,63 @@
...
+     * Generates a new, deterministic {@code ECKeyPair} instance based on the
+     * provided seed.
+     *
+     * @param seed The seed to use when deriving the key pair instance.
+     * @return A key pair that corresponds to the provided seed.
+     * @throws IllegalArgumentException if the seed is empty or a null pointer.
+     */
+    public ECKeyPair generateKeyPairFromSeed(byte[] seed) {
+        if (seed == null) {
+            throw new IllegalArgumentException("Seed must not be null");
```
talekhinezh: We've been using `Objects.requireNonNull` for null checks. Can we replace that here?

radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/ECKeyPairTest.java
```diff
@@ -1,18 +1,20 @@
...
+
 public class ECKeyPairTest {
-	@Test
-	public void decryptBadEncryptedDataWithGoodEncryptedPrivateKeyShouldThrowCryptoException() {
-		ECKeyPair keyPair = ECKeyPairGenerator.newInstance().generateKeyPair();
-		ECKeyPair privateKey = ECKeyPairGenerator.newInstance().generateKeyPair();
 
-		EncryptedPrivateKey encryptedPrivateKey = privateKey.encryptPrivateKey(keyPair.getPublicKey());
+    @Test
+    public void decrypt_bad_encrypted_data_with_good_encrypted_private_key__should_throw_CryptoException() {
```
talekhinezh: I don't believe this passes checkstyle after running `./gradlew build`. I've tried updating `checkstyle.xml` in the past but ran into problems then got lazy :) Can you either update checkstyle.xml or fix the method signature so that `./gradlew build` passes?

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/ECKeyPairGenerator.java
```diff
@@ -19,23 +20,28 @@
...
-					: SECNamedCurves.getByName("secp" + bits + "k1");
-			return new ECDomainParameters(curve.getCurve(), curve.getG(), curve.getN(), curve.getH());
-		}));
-	}
+    static {
+        install();
+    }
+
+    static void install() {
+        if (AndroidUtil.isAndroidRuntime()) {
```
talekhinezh: Can we add a check or flag here so that we don't call `install()` multiple times?

radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/RadixECKeyPairsTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.hamcrest.core.IsNot.not;
+import static org.junit.Assert.assertThat;
+
+public class RadixECKeyPairsTest {
+
+    @BeforeClass
+    public static void beforeSuite() {
+        // Ensure the BouncyCastle providers are loaded into memory
+        // (because BouncyCastle SHA-256 is used in "seed" tests).
+        ECKeyPairGenerator.newInstance();
```
talekhinezh: Is this necessary?

talekhinezh:
@laszlourszuly Just a few minor comments and yes, the change is isolated enough that I'm cool with it :) Thanks!

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/RadixECKeyPairs.java
laszlourszuly: Sure! I was aiming for consistency with the next block (checking for empty seed), but I can't rectify my choice beyond that.

radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/RadixECKeyPairsTest.java
laszlourszuly: No, not `newInstance()`, but `install()` is necessary sometimes, depending on what has been run and what hasn't before these tests are executed.

laszlourszuly: Hm... the code comment isn't too clear about the fact that the "seededKeyPair" implementation will try to retrieve the "BC" provider and it's "SHA-256" implementation. Unless we ensure that said provider exists in the java `Security` repository (by `install()`-ing it), we will get an exception and thereby failing tests.

On top of all, this only happens if no `ECKeyPairGenerator` related code has been run before these tests are executed - "random" at first glance.

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/ECKeyPairGenerator.java
laszlourszuly: Good catch! I'll look into this tomorrow (there myst be a way to check what version of bouncy castle is currently loaded in `Security`).

laszlourszuly: Fixed!

Having worked a bit closer with this particular code, I must say I don't fully see the benefit of involving `stream`'s, with all the whistles and bells that comes with it, only to add a single `Integer` mapping to a `HashMap`:

```java
DOMAINS.putAll(Stream.of(256)
        .collect(Collectors.toMap(Integer::new, (bits) -> {
                final X9ECParameters curve = bits == 256
                        ? CustomNamedCurves.getByName("secp" + bits + "k1")
                        : SECNamedCurves.getByName("secp" + bits + "k1");
                return new ECDomainParameters(curve.getCurve(), curve.getG(), curve.getN(), curve.getH());
        })));
```

vs.

```java
X9ECParameters curve = CustomNamedCurves.getByName("secp" + bits + "k1");
DOMAINS.put(256, new ECDomainParameters(curve.getCurve(), curve.getG(), curve.getN(), curve.getH());
```

laszlourszuly:
Thanks for really good feedback! I have addressed your points, but I have an additional question regarding how we populate the `DOMAINS` map in the `ECKeyPairGenerator` class.

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/ECKeyPairGenerator.java
```diff
@@ -21,21 +25,47 @@
...
-	}
+    static {
+        install();
+    }
+
+    /**
+     * Ensures the proper version of BouncyCastle is installed in the Java Security
+     * repository and that correct secpXYZk1 elliptic curves can be derived.
+     */
+    static void install() {
```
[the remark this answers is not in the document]
laszlourszuly: Good point. Fixed.

radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/ECKeyPair.java
```diff
@@ -30,7 +30,7 @@
 
 @SerializerId2("ECKEYPAIR")
 public class ECKeyPair extends SerializableObject {
-	@JsonProperty("public")
+    @JsonProperty("public")
```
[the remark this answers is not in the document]
laszlourszuly: This is the coder version of "auto-correct" mishaps when texting :)

.checkstyle_test.xml
```diff
@@ -0,0 +1,181 @@
+<?xml version="1.0"?>
```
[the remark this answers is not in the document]
laszlourszuly: Woot woot, indeed! :D

PR #121, opened 2018-12-03
Expose method for creating api with custom mappers

talekhinezh:
* Expose method in RadixApplicationAPI to allow for custom mappers
* MessageToParticles mapper more generic in order to be able to test different encryption cases

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
@@ -132,8 +132,35 @@ public class RadixApplicationAPI {
 		RadixIdentity identity,
 		RadixUniverse universe,
 		FeeMapper feeMapper,
+		Ledger ledger,
+		List<ActionToParticlesMapper> actionToParticlesMappers
+	) {
+		this.identity = identity;
+		this.universe = universe;
+
+		// TODO: Utilize class loader to discover and load these modules
+		this.uniquePropertyTranslator = new UniquePropertyTranslator();
+		this.messageActionStore = new ActionStore<>(ledger.getAtomStore(), new AtomToDecryptedMessageMapper(universe));
+		this.tokenBalanceStore = new ApplicationStore<>(ledger.getParticleStore(), new TokenBalanceReducer());
+		this.tokenTransferActionStore = new ActionStore<>(ledger.getAtomStore(), new AtomToTokenTransfersMapper(universe));
+		this.tokenStore = new ApplicationStore<>(ledger.getParticleStore(), new TokenReducer());
+
+		this.actionToParticlesMappers = actionToParticlesMappers;
+		this.feeMapper = feeMapper;
+		this.ledger = ledger;
+	}
+
+	private RadixApplicationAPI(
+		RadixIdentity identity,
+		RadixUniverse universe,
+		FeeMapper feeMapper,
 		Ledger ledger
 	) {
+		Objects.requireNonNull(identity);
+		Objects.requireNonNull(universe);
+		Objects.requireNonNull(feeMapper);
+		Objects.requireNonNull(ledger);
+
 		this.identity = identity;
 		this.universe = universe;
 
@@ -164,6 +191,7 @@ public class RadixApplicationAPI {
 
 	public static RadixApplicationAPI create(RadixIdentity identity) {
 		Objects.requireNonNull(identity);
+
 		return create(
 			identity,
 			RadixUniverse.getInstance(),
@@ -176,10 +204,29 @@ public class RadixApplicationAPI {
 		RadixUniverse universe,
 		FeeMapper feeMapper
 	) {
-		Objects.requireNonNull(identity);
-		Objects.requireNonNull(universe);
-		Objects.requireNonNull(feeMapper);
 		return new RadixApplicationAPI(identity, universe, feeMapper, universe.getLedger());
+	}
+
+	/**
+	 * For advanced use at the moment.
+	 *
+	 * Create an api based on actionToParticleMappers of own choosing.
+	 *
+	 * @param identity identity for api
+	 * @param actionToParticlesMappers the mappers to utilize for api
+	 * @return api object used to interact with ledger
+	 */
+	public static RadixApplicationAPI create(
+		RadixIdentity identity,
+		List<ActionToParticlesMapper> actionToParticlesMappers
+	) {
+		return new RadixApplicationAPI(
+			identity,
+			RadixUniverse.getInstance(),
+			new PowFeeMapper(p -> new Atom(p).getHash(), new ProofOfWorkBuilder()),
+			RadixUniverse.getInstance().getLedger(),
+			actionToParticlesMappers
+		);
 	}
 
 	/**
--- a/radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
@@ -3,10 +3,12 @@ package com.radixdlt.client.application.translate.data;
 import com.google.gson.JsonArray;
 import com.radixdlt.client.application.translate.Action;
 import com.radixdlt.client.application.translate.ActionToParticlesMapper;
+import com.radixdlt.client.atommodel.accounts.RadixAddress;
 import com.radixdlt.client.core.atoms.particles.SpunParticle;
 import com.radixdlt.client.atommodel.message.MessageParticle;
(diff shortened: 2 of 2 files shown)


radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -128,12 +128,39 @@ public Completable toCompletable() {
...
+	) {
+		this.identity = identity;
+		this.universe = universe;
+
+		// TODO: Utilize class loader to discover and load these modules
+		this.uniquePropertyTranslator = new UniquePropertyTranslator();
+		this.messageActionStore = new ActionStore<>(ledger.getAtomStore(), new AtomToDecryptedMessageMapper(universe));
+		this.tokenBalanceStore = new ApplicationStore<>(ledger.getParticleStore(), new TokenBalanceReducer());
+		this.tokenTransferActionStore = new ActionStore<>(ledger.getAtomStore(), new AtomToTokenTransfersMapper(universe));
+		this.tokenStore = new ApplicationStore<>(ledger.getParticleStore(), new TokenReducer());
```
laszlourszuly: Nitpicking, but would it make sense to move these instantiations to the static creator method instead? Usually you'd prefer to return quickly from the constructor, which can't really be guaranteed right now. And also because testing (mocking)...

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -128,12 +128,39 @@ public Completable toCompletable() {
...
+
+		this.actionToParticlesMappers = actionToParticlesMappers;
+		this.feeMapper = feeMapper;
+		this.ledger = ledger;
```
laszlourszuly: Maybe null-check these too? Especially the `actionToParticlesMapper` (which, actually, could be tested for emptiness as well) as it ultimately comes from the caller.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -128,12 +128,39 @@ public Completable toCompletable() {
...
 
+	private RadixApplicationAPI(
+		RadixIdentity identity,
+		RadixUniverse universe,
+		FeeMapper feeMapper,
+		Ledger ledger,
+		List<ActionToParticlesMapper> actionToParticlesMappers
+	) {
+		this.identity = identity;
+		this.universe = universe;
```
laszlourszuly: Would we want to null-check these? (we do it in the other constructor)

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
```diff
@@ -15,52 +17,64 @@
...
+	/**
+	 * Function describing given an action who do we encrypt a message for
+	 */
+	private final Function<SendMessageAction, Stream<ECPublicKey>> encryptionScheme;
+
 	public SendMessageToParticlesMapper(Supplier<ECKeyPair> generator) {
+		this(generator, sendMsg -> Stream.of(sendMsg.getFrom(), sendMsg.getTo()).map(RadixAddress::getPublicKey));
+	}
+
+	public SendMessageToParticlesMapper(Supplier<ECKeyPair> generator, Function<SendMessageAction, Stream<ECPublicKey>> encryptionScheme) {
```
laszlourszuly: Funny how the private member fields were JavaDoc-ed, but these public methods were not :)

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -176,12 +204,31 @@ public static RadixApplicationAPI create(
...
 
+	/**
+	 * For advanced use at the moment.
+	 *
+	 * Create an api based on actionToParticleMappers of own choosing.
+	 *
+	 * @param identity identity for api
+	 * @param actionToParticlesMappers the mappers to utilize for api
+	 * @return api object used to interact with ledger
+	 */
```
laszlourszuly: Nitpicking2: A beautifully written JavaDoc method description uses a 3rd person declarative form ("Gets this and that") rather than 2nd person imperative form ("Get this and that").

If you, just as I, get excited about JavaDoc recommendations, there is a good article about it here: https://www.oracle.com/technetwork/java/javase/index-137868.html

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
```diff
@@ -15,52 +17,64 @@
...
 public class SendMessageToParticlesMapper implements ActionToParticlesMapper {
+
+	/**
+	 * A module capable of creating new securely random ECKeyPairs
+	 */
 	private final Supplier<ECKeyPair> generator;
 
+	/**
+	 * Function describing given an action who do we encrypt a message for
+	 */
```
laszlourszuly: Nitpicking3: This description is a bit cryptic to me.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
```diff
@@ -15,52 +17,64 @@
...
+	 * If SendMessageAction is unencrypted, returns a single message particle containing the
+	 * payload data.
+	 *
+	 * If SendMessageAction is encrypted, creates a private key encrypted by both from and to
+	 * users, stores that into a message particles and then creates another message particle
+	 * with the payload encrypted by the newly created private key.
+	 *
+	 * @param action the action to map to particles
+	 * @return an observable particles of a SendMessageAction. Otherwise, an empty observable
+	 */
```
laszlourszuly: Maybe this description would benefit from one more iteration.

* It's a bit unclear what `SendMessageAction` refers to. Is it the provided `Action` argument?
* The description seems to focus a bit too much on "how" rather than "what".

There are a couple of recommendations in above resource regarding this as well ("Implementation-independence", "First sentence" etc)

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: Yes, this definitely needs to be refactored out. But I would rather have tech debt here to remind us that this does need to be figured out rather than prematurely refactoring. I think we still need a few more apps written on top of the api to inform us the correct way to refactor this.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: Yes! thanks for catching that.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
talekhinezh: It's hilarious!

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: Unfortunately, I've always been lackadaisical with javadocs which is why you're here! Cool resource and will hopefully improve on this with your guidance :)

laszlourszuly: You’re so humble! You don’t deserve an ass hole reviewer like me...

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/data/SendMessageToParticlesMapper.java
```diff
@@ -15,52 +17,79 @@
...
+	 * payload data.
+	 *
+	 * If SendMessageAction is encrypted, creates a private key encrypted by both from and to
+	 * users, stores that into a message particles and then creates another message particle
+	 * with the payload encrypted by the newly created private key.
+	 *
+	 * @param action the action to map to particles
+	 * @return observable of spunparticles to be included in an atom for a given action
+	 */
 	public Observable<SpunParticle> map(Action action) {
```
flotothemoon: Realise this hasn't changed - why is this an `Observable` instead of, say, a `List` or `Collection`. Doesn't seem like there is anything async going on there to me.

talekhinezh: Missing an @Override to show that this method is part of interface.

PR #122, opened 2018-12-04
Feature/configurable radix network

flotothemoon:
I rewrote it again to make a proper RadixPeerState, now I'm quite happy with it. Tests aren't updated yet to reflect those changes, but apart from that it's done.


radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixClientSupplier.java
```diff
@@ -0,0 +1,177 @@
...
+	 *
+	 * @param shards set of shards to find an intersection with
+	 * @return a cold single of the first matching Radix client
+	 */
+	public Single<RadixJsonRpcClient> getRadixClient(Set<Long> shards) {
+		if (shards.isEmpty()) {
+			throw new IllegalArgumentException("Shards cannot be empty to obtain a radixClient.");
+		}
+
+		List<RadixPeerFilter> expandedFilters = new ArrayList<>(this.filters);
```
talekhinezh: `this.filters` needs to be thread protected or become an unmodifiableList (probably the latter).

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixClientSupplier.java
```diff
@@ -0,0 +1,177 @@
...
+	 * @param filters The filters used to test the desirability of peers
+	 */
+	public RadixClientSupplier(RadixNetwork network, RadixPeerSelector selector, List<RadixPeerFilter> filters) {
+		Objects.requireNonNull(network, "network is required");
+		Objects.requireNonNull(selector, "selector is required");
+		Objects.requireNonNull(filters, "filters is required");
+
+		this.network = network;
+		this.selector = selector;
+		this.filters = filters;
```
talekhinezh: `filters` here should probably be copied.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -77,19 +73,19 @@ public RadixJsonRpcClient(WebSocketClient wsClient) {
...
-		} else {
-			this.serverApiVersion = jsonRpcCall("Api.getVersion")
-				.map(result -> result.getAsJsonObject().get("version").getAsInt())
-				.cache();
-		}
+		this.serverApiVersion = SingleSubject.create();
+		this.serverApiVersion.onSuccess(API_VERSION);
+		jsonRpcCall("Api.getVersion")
+			.map(result -> result.getAsJsonObject().get("version").getAsInt())
+			.subscribe(serverApiVersion::onSuccess, e -> LOGGER.error("Error while requesting Api.getVersion: " + e, e));
```
talekhinezh: This looks wrong to me. `onSuccess` should only ever be called once and is called twice here no?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -77,19 +73,19 @@ public RadixJsonRpcClient(WebSocketClient wsClient) {
...
 
 		Serialization serialization = Serialize.getInstance();
-		this.universeConfig = jsonRpcCall("Universe.getUniverse")
+
+		this.universeConfig = SingleSubject.create();
+		jsonRpcCall("Universe.getUniverse")
 			.map(element -> GsonJson.getInstance().stringFromGson(element))
 			.map(result -> serialization.fromJson(result, RadixUniverseConfig.class))
-			.cache();
+			.subscribe(universeConfig::onSuccess);
```
talekhinezh: I don't think we want JsonRpcClient making the subscribe calls. These should be called by an outside class. It feels wrong that instantiating a class automatically makes network calls.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -99,7 +95,7 @@ public String getLocation() {
 		return wsClient.getEndpoint().url().toString();
 	}
 
-	public Observable<RadixClientStatus> getStatus() {
+	public BehaviorSubject<RadixClientStatus> getStatus() {
```
talekhinezh: We shouldn't be exposing Subjects to callers of this method as they will then be able to call `onNext`, `onSuccess` etc. which totally breaks contract.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -172,20 +168,28 @@ public boolean tryClose() {
...
 
-	public Single<Integer> getAPIVersion() {
-		return serverApiVersion;
+	public Optional<Integer> getAPIVersion() {
+		return Optional.ofNullable(serverApiVersion.getValue());
 	}
 
-	public Single<Boolean> checkAPIVersion() {
-		return this.getAPIVersion().map(API_VERSION::equals);
+	public SingleSubject<Integer> apiVersion() {
```
talekhinezh: Same thing here, remove these Subject returns.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkState.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Current state in time of a {@link RadixNetwork}
+ */
+public class RadixNetworkState {
+	public final Map<RadixPeer, RadixPeerState> peers;
+
+	public RadixNetworkState(Map<RadixPeer, RadixPeerState> peers) {
+		Objects.requireNonNull(peers, "peers is required");
+
+		this.peers = Collections.unmodifiableMap(peers);
```
talekhinezh: Might want to copy `peers` as well.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixPeer.java
```diff
@@ -57,8 +70,12 @@ public RadixPeer data(NodeRunnerData data) {
 		return this;
 	}
 
-	public Maybe<RadixPeer> servesShards(Set<Long> shards) {
-		return data.map(d -> d.getShards().intersects(shards)).flatMapMaybe(intersects -> intersects ? Maybe.just(this) : Maybe.empty());
+	public SingleSubject<NodeRunnerData> getData() {
```
talekhinezh: Remove Subject

radixdlt-java/src/main/java/com/radixdlt/client/core/network/WebSocketClient.java
```diff
@@ -51,7 +48,7 @@ public Request getEndpoint() {
 		return endpoint;
 	}
 
-	public Observable<RadixClientStatus> getStatus() {
+	public BehaviorSubject<RadixClientStatus> getStatus() {
```
talekhinezh: Remove Subject

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixClientSupplier.java
flotothemoon: Now is copied into immutable list

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
flotothemoon: As discussed, this was a hack as the core currently doesn't support Api.getVersion. I've improved the hack until the core fixes it to return a compatible API version on error.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
flotothemoon: Agreed, fixed.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
flotothemoon: Ha, yeah, I noticed that as well while doing the tests. "Why can I call onNext? Ohhhhhh.." :D

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -77,19 +72,23 @@ public RadixJsonRpcClient(WebSocketClient wsClient) {
...
-		} else {
-			this.serverApiVersion = jsonRpcCall("Api.getVersion")
-				.map(result -> result.getAsJsonObject().get("version").getAsInt())
-				.cache();
-		}
+		this.serverApiVersion = SingleSubject.create();
+		this.universeConfig = SingleSubject.create();
+	}
 
+	void fetchClientInformation() {
```
talekhinezh: This function is still calling subscribe on it's own when really that should be done with a returned Observable or Single or Completable. Otherwise you lose the ability to dispose or catch errors.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -77,19 +72,23 @@ public RadixJsonRpcClient(WebSocketClient wsClient) {
...
 		Serialization serialization = Serialize.getInstance();
-		this.universeConfig = jsonRpcCall("Universe.getUniverse")
+		jsonRpcCall("Api.getVersion")
+				.map(result -> result.getAsJsonObject().get("version").getAsInt())
+				.onErrorReturn(e -> {
+					LOGGER.error(String.format("Error while requesting Api.getVersion: %s", e));
+					return API_VERSION; // assume api version matches for now until fixed in core
+				})
+				.subscribe(serverApiVersion::onSuccess);
```
talekhinezh: Still not sure why original method was removed. Doing it this way makes you lose error information in your serverApiVersion subject. Subjects should be used with care in RX and only when necessary as they add a lot of state management: https://medium.com/upday-devs/subjects-neither-good-nor-evil-8384ad23a2ef

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -77,19 +72,23 @@ public RadixJsonRpcClient(WebSocketClient wsClient) {
...
+		jsonRpcCall("Universe.getUniverse")
 			.map(element -> GsonJson.getInstance().stringFromGson(element))
 			.map(result -> serialization.fromJson(result, RadixUniverseConfig.class))
-			.cache();
+			.subscribe(universeConfig::onSuccess);
```
talekhinezh: Same comment as above.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -99,7 +98,11 @@ public String getLocation() {
 		return wsClient.getEndpoint().url().toString();
 	}
 
-	public Observable<RadixClientStatus> getStatus() {
+	public Observable<RadixClientStatus> status() {
+		return wsClient.status();
+	}
+
+	public Optional<RadixClientStatus> getStatus() {
 		return wsClient.getStatus();
```
talekhinezh: Code smell here in that we need both synchronous and asynchronous versions. Something must be wrong somewhere...If this is being used for the pure NetworkState function this `status` should be `flatMapped` somewhere...

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
```diff
@@ -172,20 +175,28 @@ public boolean tryClose() {
...
+	}
+
+	public Single<Integer> apiVersion() {
+		return this.serverApiVersion;
 	}
 
-	public Single<Boolean> checkAPIVersion() {
-		return this.getAPIVersion().map(API_VERSION::equals);
+	public boolean checkAPIVersion() {
+		return this.getAPIVersion().map(API_VERSION::equals).orElse(false);
```
talekhinezh: Again some code smell due to async + sync versions.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
```diff
@@ -25,28 +24,41 @@
 	/**
 	 * Hot observable which updates subscribers of new connection events
 	 */
-	private final ConnectableObservable<SimpleImmutableEntry<String, RadixClientStatus>> statusUpdates;
-
+	private final ConnectableObservable<SimpleImmutableEntry<RadixPeer, RadixClientStatus>> statusUpdates;
```
talekhinezh: RadixPeer isn't immutable so weird to have it as a key.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
```diff
@@ -25,28 +24,41 @@
...
+	private final Observable<RadixNetworkState> networkState;
 
 	public RadixNetwork(PeerDiscovery peerDiscovery) {
 		Objects.requireNonNull(peerDiscovery);
 
 		this.peers = peerDiscovery.findPeers()
 			.retryWhen(new IncreasingRetryTimer(WebSocketException.class))
-			.doOnNext(peer -> LOGGER.info("Added to peer list: " + peer.getLocation()))
+			.doOnNext(peer -> {
+				peer.getRadixClient().fetchClientInformation();
```
talekhinezh: This opens a connection to every peer but I don't believe it ever closes the connection. Is that right?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
```diff
@@ -25,28 +24,41 @@
...
+				.scan(new RadixNetworkState(Collections.emptyMap()), (previousState, update) -> {
+			LinkedHashMap<RadixPeer, RadixPeerState> currentPeers = new LinkedHashMap<>(previousState.getPeers());
+			currentPeers.put(update.getKey(), update.getValue());
+
+			return new RadixNetworkState(currentPeers);
+		});
 	}
 
-	public Observable<SimpleImmutableEntry<String, RadixClientStatus>> connectAndGetStatusUpdates() {
+	public Observable<SimpleImmutableEntry<RadixPeer, RadixClientStatus>> connectAndGetStatusUpdates() {
```
talekhinezh: Same issue here, weird to have RadixPeer as key when RadixPeer is mutable and has state.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkState.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Current state in time of a {@link RadixNetwork}
+ */
+public class RadixNetworkState {
+	private final Map<RadixPeer, RadixPeerState> peers;
+
+	public RadixNetworkState(Map<RadixPeer, RadixPeerState> peers) {
```
talekhinezh: In future, using the builder pattern would probably be easiest way to construct this so you wouldn't need to pass maps around.

talekhinezh: But no worries for now.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
flotothemoon: This isn't strictly a key as it's just a single entry. For clarity I've changed it to use Pair instead.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkState.java
flotothemoon: That's a good idea yeah. As you say, not necessary right now.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
flotothemoon: After further consideration I've changed it to just return Observable<RadixPeerState>, that seems cleaner. There is technically a similar issue in RadixNetworkState where RadixPeer is used as a key, so I've changed that to an identity map to be safe (as we do want an instance of the actual RadixPeer there, otherwise it's kind of useless - could change so that you can look up peers in network via their peer state, but for now we're good I think).

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixJsonRpcClient.java
flotothemoon: Ignore all that, has been removed.

PR #123, opened 2018-12-05
Replace JUnit4 with JUnit5

MarcRubio:
I have replaced JUnit4 with JUnit5 with some new testCompile dependencies which keep it backwards compatible.

In order to quickly compare the difference I have duplicated RadixECKeyPairsTest naming it RadixECKeyPairsTestJUnit5

To see what annotations are available and how they compare to JUnit4 please see https://junit.org/junit5/docs/current/user-guide/#writing-tests-annotations

JUnit5 offers a bunch more of powerful features which can significantly reduce then number of code lines and keep things much neater with its annotations. 

A feature to consider is the possibility of using the Nested annotation which is basically using inner classes where Test methods use a common variable. This allows to shorten the @DisplayName even further since the "Given" part would be common and a @DisplayName can be placed on top of the inner class.

Also, in case this is missed, when tests are run the @DisplayName is shown instead of the method signatures.

### DIFF
--- a/radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/RadixECKeyPairsTestJUnit5.java
+++ b/radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/RadixECKeyPairsTestJUnit5.java
@@ -1,0 +1,52 @@
+package com.radixdlt.client.core.crypto;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.IsNot.not;
+import static org.junit.Assert.assertThat;
+
+public class RadixECKeyPairsTestJUnit5 {
+
+    @BeforeEach
+    public void beforeSuite() {
+        // Ensure the BouncyCastle providers are loaded into memory
+        // (because BouncyCastle SHA-256 is used in "seed" tests).
+        ECKeyPairGenerator.install();
+    }
+
+    @Test
+    @DisplayName("Given privateKey1 and privateKey2 When we compare them Then they should be different")
+    public void generateAndCompareTwoPrivateKeys() {
+        byte[] privateKey1 = ECKeyPairGenerator
+                .newInstance()
+                .generateKeyPair()
+                .getPrivateKey();
+
+        byte[] privateKey2 = ECKeyPairGenerator
+                .newInstance()
+                .generateKeyPair()
+                .getPrivateKey();
+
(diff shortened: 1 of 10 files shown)


radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/RadixECKeyPairsTestJUnit5.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @BeforeEach
+    public void beforeSuite() {
+        // Ensure the BouncyCastle providers are loaded into memory
+        // (because BouncyCastle SHA-256 is used in "seed" tests).
+        ECKeyPairGenerator.install();
+    }
+
+    @Test
+    @DisplayName("Given privateKey1 and privateKey2 When we compare them Then they should be different")
```
flotothemoon: Can we break these up over multiple lines?

MarcRubio: I don't believe so but come one... you should be happy, look at all those spaces you can add!

MarcRubio: Just a note though, to break it up, an inner class using @Nested annotation could be used and you can place the @DisplayName("Given privateKey1 and privateKey2") there on top of the inner class and keeping @DisplayName("When we compare them Then they should be different") on the method inside the inner class which will make it shorter and look nicer.

This is used when there are multiple methods using the same "Given". In this case though, there is only one method so not really necessary but would make it look neater for sure.

mpsandiford:
Which issue is this part of?  I don't remember this being an agreed task for the sprint.

radixdlt-java/src/test/java/com/radixdlt/client/core/crypto/RadixECKeyPairsTestJUnit5.java
mpsandiford: It's possible to use string concatenation in annotations in Java, so could be something like:
```
@DisplayName(
    "Given privateKey1 and privateKey2 " +
    "When we compare them " +
    "Then they should be different")
```

laszlourszuly:
@MarcRubio is this something that should be merged? If not, we should consider closing it, otherwise a rebase and merge perhaps?

MarcRubio:
I agree about doing something. This change was a non breaking and the natural upgrade to JUnit4 while keeping things compatible across the board. But since cucumber framework has been adopted for testing I think we can close it as this can be easily added in the future.

laszlourszuly:
Closing as of ack in previous comment.

PR #126, opened 2018-12-12
Implement auto-disconnect when all Observables are disposed

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -355,6 +356,62 @@ public TokenClassReference getNativeTokenRef() {
...
+				: Observable.never();
+		Observable<AtomObservation> auto = atomsPulled.publish()
+			.refCount(2);
+		Disposable d = auto.subscribe();
+
+		return this.getStore(stateClass).getState(address)
+			.map(stateClass::cast)
+			.publish()
+			.autoConnect()
+			.doOnSubscribe(disposable -> auto.subscribe().dispose())
```
flotothemoon: I'm a bit fuzzy on how the logic here works. Why does it subscribe and immediately dispose of `auto`?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -355,6 +356,62 @@ public TokenClassReference getNativeTokenRef() {
...
+			.doOnError(e -> d.dispose())
```
flotothemoon: We seem to lose error information here.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -363,9 +420,7 @@ public TokenClassReference getNativeTokenRef() {
 	 * @return a hot observable of the latest state of token classes
 	 */
 	public Observable<TokenClassesState> getTokenClasses(RadixAddress address) {
-		pull(address);
-
-		return this.getStore(TokenClassesState.class).getState(address).map(s -> (TokenClassesState) s);
+		return getState(TokenClassesState.class, address);
```
flotothemoon: The `this` has `this`-appeared. Ha. Ha.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ApplicationStore.java
```diff
@@ -24,8 +25,7 @@ public ApplicationStore(
 			particleStore.getParticles(address)
 				.scanWith(reducer::initialState, reducer::reduce)
 				.debounce(1000, TimeUnit.MILLISECONDS)
-				.replay(1)
-				.autoConnect()
+				.compose(ReplayingShare.instance())
```
flotothemoon: Do we really need to include an entire library for this line?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
```diff
@@ -51,6 +53,18 @@ public RadixNetwork(PeerDiscovery peerDiscovery) {
 		return this.getStatusUpdates();
 	}
 
+	public Observable<Map<String, RadixClientStatus>> getNetworkState() {
```
flotothemoon: I suppose this will have to be updated when the `configurable-network` branch has been merged (hopefully today)? Or does this need some other functionality that the `configurable-network` branch's `RadixNetworkState` can't provide?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: This is done in order to only subscribe to the AtomPull only when the observable we are returning is subscribed to and to be able to dispose of the AtomPull when the returned observable is diposed of.

To be fair this may not be the best of implementing this but without more significant refactoring this was the best way I could figure out how to implement for now.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: Don't need the error though for this lambda, it just gets propagated downstream to the correct handler.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ApplicationStore.java
talekhinezh: Yes, the whole library is actually just this one line. It's a super useful construct which can't be replicated easily in RXJava.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetwork.java
talekhinezh: Yes, this is actually more temporary in order to be able to prove some things in the Regression/Acceptance tests.

PR #127, opened 2018-12-12
Fix wrapped shard range logic

talekhinezh:


radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/ShardsTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class ShardsTest {
+	@Test
+	public void inclusionTest() {
+		Shards shards = Shards.range(1, 1);
+		assertThat(shards.contains(1)).isTrue();
+		assertThat(shards.contains(2)).isFalse();
+		assertThat(shards.intersects(Arrays.asList(1L, 2L, 3L, 4L, 5L))).isTrue();
+		assertThat(shards.intersects(Arrays.asList(2L, 3L, 4L, 5L))).isFalse();
+	}
```
laszlourszuly: I think the generic test method name, the anonymous magic numbers and the mix of positive and negative assertions requires the reader to have deep domain knowledge to understand what's really being tested here.

I may be totally wrong (lacking domain knowledge :wink:) but could the above test be split into two isolated tests, like below? The example below would in my opinion be a bit more user friendly mostly because of the human readable naming.

EXAMPLE:
```java
@Test
public void when_checking_if_included_shard_is_in_shard_space__intersection_can_be_detected() {
    long includedShard = 1L;
    List<Long> someShards = Arrays.asList(includedShard, 2L, 3L, 4L, 5L);
    Shard shardSpace = Shards.range(1L, 1L);

    assertThat(shardSpace.contains(includedShard)).isTrue();
    assertThat(shardSpace.intersects(someShards)).isTrue();
}

@Test
public void when_checking_if_excluded_shard_is_in_shard_space__intersection_can_not_be_detected() {
    long excludedShard = 2L;
    List<Long> someShards = Arrays.asList(excludedShard, 3L, 4L, 5L);
    Shard shardSpace = Shards.range(1L, 1L);

    assertThat(shardSpace.contains(excludedShard)).isFalse();
    assertThat(shardSpace.intersects(someShards)).isFalse();
}
```

flotothemoon:
Looks good to me, though I agree with @laszlourszuly a bit on the magic numbers. A simple comment would suffice in the tests I think.

talekhinezh:
@laszlourszuly @flotothemoon Thanks guys, fixed based on comments and will merge now.

PR #129, opened 2018-12-14
Implement RRI

talekhinezh:
Replace NonFungibleQuark with IdentifiableQuark


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenClassReference.java
```diff
@@ -1,22 +1,17 @@
-package com.radixdlt.client.atommodel.tokens;
+package com.radixdlt.client.application.translate.tokens;
```
flotothemoon: I get why `RadixResourceIdentifier` is in application, but I don't understand why `TokenClassReference` is in `application`, not `atommodel`. It sounds like an `atommodel`-level aspect to me.

radixdlt-java/src/main/java/org/radix/serialization2/mapper/JacksonCborMapper.java
```diff
@@ -76,6 +78,11 @@ public static JacksonCborMapper create(SerializerIds idLookup, FilterProvider fi
 			JacksonCodecConstants.U20_VALUE,
 			UInt256::toByteArray
 		));
+		cborModule.addSerializer(RadixResourceIdentifer.class, new JacksonCborObjectBytesSerializer<>(
+			RadixResourceIdentifer.class,
+			JacksonCodecConstants.RRI_VALUE,
+			id -> id.toString().getBytes(RadixConstants.STANDARD_CHARSET)
```
flotothemoon: Same comment as on `RadixCore`, would like a `to` and `from` `byte[]` function in `RadixResourceIdentifier`.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenClassReference.java
talekhinezh: RRI is actually the only thing that is `atommodel` level at the moment since that is the input into quarks. `TokenClassReference` on the other hand isn't used in the atom model and is just a helper to map address + symbol to address + type + name.

PR #132, opened 2019-01-08
Feature/refactor network

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
```diff
@@ -66,28 +78,43 @@
 	 * Initializes the default universe with a Peer Discovery mechanism.
```
flotothemoon: "with a Peer Discovery mechanism" should be changed to "from a set of seed nodes" or similar

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/RadixJsonRpcMethodsEpic.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@Override
+	public Observable<RadixNodeAction> epic(Observable<RadixNodeAction> actions, Observable<RadixNetworkState> networkState) {
+		return actions
+				.filter(a -> a instanceof JsonRpcMethodAction)
+				.map(JsonRpcMethodAction.class::cast)
+				.flatMapMaybe(a -> {
+					final WebSocketClient ws = webSockets.get(a.getNode());
+					return ws.getState()
+						.filter(s -> s.equals(WebSocketStatus.CONNECTED))
```
flotothemoon: Could be simplified to `WebSocketStatus.CONNECTED::equals`

radixdlt-java/src/main/java/com/radixdlt/client/core/network/jsonrpc/RadixJsonRpcClient.java
```diff
@@ -62,106 +82,63 @@
...
+				.firstOrError()
+				.map(msg -> {
+					final JsonObject jsonResponse = msg.getAsJsonObject();
+					return new JsonRpcResponse(!jsonResponse.has("error"), jsonResponse);
+				})
+				.subscribe(emitter::onSuccess);
+
+			boolean sendSuccess = channel.sendMessage(GsonJson.getInstance().stringFromGson(requestObject));
+			if (!sendSuccess) {
+				emitter.onError(new RuntimeException("Could not send message: " + method + " " + params));
```
flotothemoon: Is the error information from emitter discarded here? Is this intentional?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/jsonrpc/RadixJsonRpcClient.java
```diff
@@ -239,6 +219,54 @@ protected boolean tryClose() {
 		return this.jsonRpcSubscribe(method, new JsonObject(), notificationMethod);
 	}
 
+	public Observable<JsonObject> observeNotifications(String notificationMethod, String subscriberId) {
```
flotothemoon: `subscriberId` is the first argument for all the other methods, so it should be first here as well I think. Consistency :)

radixdlt-java/src/main/java/com/radixdlt/client/core/network/reducers/RadixNetwork.java
```diff
@@ -0,0 +1,53 @@
...
+ * Reducer which controls state transitions as new network actions occur.
+ * Explicitly does not contain state and should be maintained as a pure function.
+ */
+public final class RadixNetwork {
+	public RadixNetwork() {
+	}
+
+	public RadixNetworkState reduce(RadixNetworkState state, RadixNodeAction action) {
+		Map<RadixNode, RadixNodeState> newMap = null;
+		if (action instanceof GetNodeDataResultAction) {
```
flotothemoon: Oh this is really cool. 
I think we should consider splitting this up into multiple methods, as all the `if (action instanceof ...)` seem to be independent and repetitive.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/selector/CompatibleApiVersionFilter.java
```diff
@@ -13,7 +13,7 @@ public CompatibleApiVersionFilter(int version) {
 	}
 
 	@Override
-	public boolean test(RadixPeerState peerState) {
+	public boolean test(RadixNodeState peerState) {
```
flotothemoon: `peerState` -> `nodeState`

radixdlt-java/src/main/java/com/radixdlt/client/core/network/selector/ConnectionAliveFilter.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.radixdlt.client.core.network.websocket.WebSocketStatus;
+import com.radixdlt.client.core.network.RadixNodeState;
+
+/**
+ * A connection status filter that filters out inactive peers
+ */
+public class ConnectionAliveFilter implements RadixPeerFilter {
+	@Override
+	public boolean test(RadixNodeState peerState) {
```
flotothemoon: All of these should be called `nodeState` instead

radixdlt-java/src/main/java/com/radixdlt/client/core/network/selector/ShardFilter.java
```diff
@@ -19,7 +19,7 @@ public ShardFilter(Set<Long> shards) {
 	}
 
 	@Override
-	public boolean test(RadixPeerState peerState) {
+	public boolean test(RadixNodeState peerState) {
```
flotothemoon: `peerState` -> `nodeState`

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	public void connect() {
+		synchronized (lock) {
+			switch (state.getValue()) {
+				case DISCONNECTED:
+				case FAILED:
+					WebSocketClient.this.state.onNext(WebSocketStatus.CONNECTING);
+					this.webSocket = this.connectWebSocket();
+					return;
+				case CONNECTING:
```
flotothemoon: Why do we need to spell out the other cases (`CONNECTING`, `CONNECTED`, `CLOSING`) here? Verbosity?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
```diff
@@ -0,0 +1,140 @@
...
+import okhttp3.WebSocketListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Manages the state for a single websocket.
+ */
+public class WebSocketClient implements PersistentChannel {
+	private static final Logger LOGGER = LoggerFactory.getLogger(WebSocketClient.class);
+	private final Object lock = new Object();
```
flotothemoon: I think it should be LOCK since it's effectively a constant. Martin pointed this out in an earlier PR and it should apply here as well

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/SubmitAtomEpic.java
```diff
@@ -0,0 +1,77 @@
...
+				if (nodeUpdate.getState().equals(NodeAtomSubmissionState.RECEIVED)) {
+					return SubmitAtomReceivedAction.of(request.getUuid(), request.getAtom(), node);
+				} else {
+					return SubmitAtomResultAction.fromUpdate(request.getUuid(), request.getAtom(), node, nodeUpdate);
+				}
+			})
+			.<RadixNodeAction>startWith(SubmitAtomSendAction.of(request.getUuid(), request.getAtom(), node))
+			.retryWhen(new IncreasingRetryTimer(WebSocketException.class))
+			// TODO: Better way of cleanup?
+			.doFinally(() -> Observable.timer(5, TimeUnit.SECONDS).subscribe(t -> ws.close()));
```
flotothemoon: Why 5 seconds? I can feel the magic ;)

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/RadixJsonRpcAutoCloseEpic.java
```diff
@@ -0,0 +1,31 @@
...
+	public RadixJsonRpcAutoCloseEpic(WebSockets webSockets) {
+		this.webSockets = webSockets;
+	}
+
+	@Override
+	public Observable<RadixNodeAction> epic(Observable<RadixNodeAction> actions, Observable<RadixNetworkState> networkState) {
+		return
+			actions
+				.filter(a -> a instanceof JsonRpcMethodAction)
+				.delay(5, TimeUnit.SECONDS)
```
flotothemoon: Magic 5 again. Should this perhaps be in some configuration file / class instead?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/FindANodeEpic.java
```diff
@@ -0,0 +1,103 @@
...
+					.replay(1)
+					.autoConnect(2);
+
+				// Try and connect if there are no nodes
+				Observable<RadixNodeAction> newConnections = syncNetState
+					.zipWith(connectedNodes.takeWhile(List::isEmpty), (s, n) -> s)
+					.flatMapMaybe(state -> findConnection(a.getShards(), state));
+
+				Observable<RadixNodeAction> selectedNode = connectedNodes
+					.filter(viablePeerList -> !viablePeerList.isEmpty())
```
flotothemoon: Where did the filters go? I don't see them used anywhere, but they're still here. Previously they were applied here

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/FetchAtomsObservationAction.java
```diff
@@ -0,0 +1,66 @@
...
+import com.radixdlt.client.atommodel.accounts.RadixAddress;
+import com.radixdlt.client.core.atoms.AtomObservation;
+import com.radixdlt.client.core.network.RadixNode;
+import java.util.Objects;
+
+/**
+ * A dispatchable fetch atoms action which represents an atom observed event from a specific node for an atom fetch flow.
+ */
+public final class FetchAtomsObservationAction implements FetchAtomsAction {
+	private final String uuid;
```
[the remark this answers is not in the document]
talekhinezh: No inheritance please! :) But really inheritance via `extend` should only be used for exceptional cases as most things should be done via composition as it avoids coupling a lot of nasty issues down the line. The "extra" code that comes from re-implementing an interface is MUCH cheaper in complexity management compared to using a common base class.  https://codeburst.io/inheritance-is-evil-stop-using-it-6c4f1caf5117 summarizes my view.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/FindANodeEpic.java
```diff
@@ -0,0 +1,103 @@
...
+			.filter(entry -> entry.getValue().getShards().map(s -> s.intersects(shards)).orElse(false))
+			.map(Map.Entry::getKey)
+			.collect(Collectors.toList());
+	}
+
+	@Override
+	public Observable<RadixNodeAction> epic(Observable<RadixNodeAction> actions, Observable<RadixNetworkState> stateObservable) {
+		return actions
+			.filter(a -> a instanceof FindANodeRequestAction)
+			.map(FindANodeRequestAction.class::cast)
```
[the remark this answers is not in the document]
talekhinezh: omg, you're right! How did I miss this?!

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/FindANodeEpic.java
talekhinezh: Yes, kept them for reference so that they can be implemented later. Removed their use as they seem to expose a limiting interface which doesn't allow a filter to make decisions based on time.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/RadixJsonRpcAutoCloseEpic.java
talekhinezh: True, perhaps just a `static final` in this class for now as I think the way this is handled can still be improved and shouldn't escape the boundaries of the class for now.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/reducers/RadixNetwork.java
talekhinezh: I actually think keeping it all in one method is cleaner as each transition is a small amount of logic and you need to do a lot less code jumping. Could perhaps refactor later if it gets too unwieldy.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
talekhinezh: I think it should be `LOCK` only if it's static no? The lock in this case is not shared between instances.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
talekhinezh: Just keeps state machine clear.

radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
flotothemoon: Keyword being *can*.
Not very important anyway, but I do think it should be "from a set of seed nodes", since the seeds don't necessarily have to come from a peer discovery mechanism.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -0,0 +1,143 @@
...
+	 *
+	 * @param atom atom to submit into the ledger
+	 * @return Observable emitting status updates to submission
+	 */
+	@Override
+	public Observable<SubmitAtomAction> submitAtom(Atom atom) {
+		SubmitAtomAction initialAction = SubmitAtomRequestAction.newRequest(atom);
+
+		Observable<SubmitAtomAction> status = nodeActions
+			.filter(a -> a instanceof SubmitAtomAction)
```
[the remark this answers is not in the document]
flotothemoon: Hmm, it makes sense but seems non-obvious to me. I think a healthy dose of documentation could resolve a lot of potential confusion regarding the Java library, particularly as I think most users won't already be familiar with Epics, Flows, ..

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNodeState.java
```diff
@@ -0,0 +1,87 @@
...
+import com.radixdlt.client.core.atoms.Shards;
+import com.radixdlt.client.core.network.jsonrpc.NodeRunnerData;
+import com.radixdlt.client.core.network.websocket.WebSocketStatus;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Immutable state at a certain point in time of a {@link RadixNode}
+ */
+public class RadixNodeState {
```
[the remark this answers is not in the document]
flotothemoon: ha ok then, not critical anyway

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/FindANodeEpic.java
flotothemoon: Sounds reasonable, what are you thinking of that will integrate time as well?

talekhinezh: e.g. selecting a node which has been online for atleast X amount of time or has not dropped out in the past minute.

radixdlt-java/src/main/java/com/radixdlt/client/core/RadixUniverse.java
talekhinezh: fair enough

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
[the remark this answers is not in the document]
talekhinezh: good point.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/FetchAtomsObservationAction.java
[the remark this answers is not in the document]
talekhinezh: I'm gonna keep pushing this because it's very important :)

The idea that inheritance should most usually be avoided (especially for java) isn't just a matter of preference but is a widely accepted view:

https://hackernoon.com/inheritance-based-on-internal-structure-is-evil-7474cc8e64dc

https://www.quora.com/Is-inheritance-bad-practice-in-OOP-Many-places-that-teach-design-patterns-say-to-opt-for-composition-over-inheritance-but-what-about-when-multiple-classes-share-logic-from-an-abstract-class-such-as-in-the-Template-Method-design-pattern

That's the point of composition, every implementation can store it how it wants. The reason to store it in every single implementation is for decoupling which keeps every implementation modular and independent from each other. Once you add a base class, now every class is dependent on the same class and any modifications to the base class will have repercussions on all classes extending it which is a mess when code gets more and more complex.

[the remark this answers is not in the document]
flotothemoon: > Dogs don’t inherit from Trees just because you want a bark() method. Keep that in mind.
Haha. Nice. 

Interesting, you got me convinced for the most part. I like composition, that's why I wanted to have Checks in the validation logic. And then again, I also implemented some of the validation structure through inheritance. How would that be improved in a composable manner? The subclasses in my validation implementation are basically just typedefs for convenience. 

We should have a discussion what we want Core to look like in a perfect world, so that we can prioritise our refactors and feature implementations properly. 

What about stuff like Quarks, Particles? I don't see any issues with having a base Particle and Quark class and more specific implementations.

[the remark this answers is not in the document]
flotothemoon: Also I love how the answer right below the one you linked me is all for inheritance again. Widely accepted indeed ;)

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
```diff
@@ -0,0 +1,140 @@
...
+					return;
+			}
+		}
+	}
+
+	public Observable<String> getMessages() {
+		return messages;
+	}
+
+	public Observable<WebSocketStatus> getState() {
```
mpsandiford: This seems to be exposing too much of the internals, as callers could do something like `getState().onNext(someRandomState)` and break all the synchronization/mutexing that is happening in this class.

Better would be to expose just the functionality required, perhaps doOnNextState.

radixdlt-java/src/test/java/com/radixdlt/client/core/network/actions/FetchAtomsCancelActionTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+
+import com.radixdlt.client.atommodel.accounts.RadixAddress;
+import org.junit.Test;
+
+public class FetchAtomsCancelActionTest {
+	@Test
+	public void when_initializing_with_null__npe_should_be_thrown() {
```
mpsandiford: Did we discuss using `_is_thrown` and `_is_not_thrown` for these?

radixdlt-java/src/test/java/com/radixdlt/client/core/network/epics/FindANodeEpicTest.java
```diff
@@ -0,0 +1,208 @@
...
+
+		testObserver.awaitCount(4);
+		testObserver.assertValueAt(0, u -> u instanceof ConnectWebSocketAction && u.getNode().equals(badPeer));
+		testObserver.assertValueAt(1, u -> u instanceof ConnectWebSocketAction && u.getNode().equals(goodPeer));
+		testObserver.assertValueAt(2, u -> u instanceof CloseWebSocketAction && u.getNode().equals(badPeer));
+		testObserver.assertValueAt(3, u -> ((FindANodeResultAction) u).getRequest().equals(request) && u.getNode().equals(goodPeer));
+	}
+
+	@Test
+	public void when_first_node_fails__then_seconde_node_should_connect() {
```
mpsandiford: Minor typo.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/FindANodeResultAction.java
```diff
@@ -0,0 +1,43 @@
+package com.radixdlt.client.core.network.actions;
+
+import com.radixdlt.client.core.network.RadixNode;
+import com.radixdlt.client.core.network.RadixNodeAction;
+import java.util.Objects;
+
+/**
+ * The result of a FindANodeRequestAction.
```
mpsandiford: Should possibly be "{@link FindANodeRequestAction}" to make it easier for readers.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
```diff
@@ -0,0 +1,140 @@
...
+				@Override
+				public void onMessage(WebSocket webSocket, String message) {
+					synchronized (lock) {
+						messages.onNext(message);
+					}
+				}
+
+				@Override
+				public void onClosing(WebSocket webSocket, int code, String reason) {
+					webSocket.close(1000, null);
```
mpsandiford: Should state not change here, given that we have just closed the socket?

radixdlt-java/src/test/java/com/radixdlt/client/core/network/actions/FetchAtomsCancelActionTest.java
flotothemoon: We did, and we decided on _exception_is_thrown and _exception_is_not_thrown variants.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/jsonrpc/RadixJsonRpcClient.java
talekhinezh: Good catch

talekhinezh: wait nvm bad catch, there is no error information as it's just a call

radixdlt-java/src/main/java/com/radixdlt/client/core/network/jsonrpc/RadixJsonRpcClient.java
talekhinezh: Mmm this is actually a lower level method which exposes a way to specify the notification method string whereas the others specify it for the caller. It should be `notificationMethod` first as it is higher in the hierarchy within the json rpc method object.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkState.java
```diff
@@ -1,24 +1,32 @@
...
+	private final Map<RadixNode, RadixNodeState> nodes;
 
-	public RadixNetworkState(Map<RadixPeer, RadixPeerState> peers) {
-		Objects.requireNonNull(peers, "peers is required");
+	public RadixNetworkState(Map<RadixNode, RadixNodeState> nodes) {
+		Objects.requireNonNull(nodes, "nodes is required");
 
-		// use identity map as peer is not immutable, but we want the convenience of one state per peer instance
-		this.peers = Collections.unmodifiableMap(new IdentityHashMap<>(peers));
+		this.nodes = Collections.unmodifiableMap(nodes);
```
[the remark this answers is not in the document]
talekhinezh: This was mainly meant to protect the class from callers of `getNodes()` but yes doesn't protect against the map which initializes the object. Will fix.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
```diff
@@ -0,0 +1,140 @@
...
+	private final Function<WebSocketListener, WebSocket> websocketFactory;
+	private final PublishSubject<String> messages = PublishSubject.create();
+	private WebSocket webSocket;
+
+	public WebSocketClient(Function<WebSocketListener, WebSocket> websocketFactory) {
+		this.websocketFactory = websocketFactory;
+		this.state
+			.filter(state -> state.equals(WebSocketStatus.FAILED))
+			.debounce(1, TimeUnit.MINUTES)
+			.subscribe(i -> {
```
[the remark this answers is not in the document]
talekhinezh: I assume you mean "Why aren't". But yes, this is a good catch.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -0,0 +1,143 @@
...
+			}
+
+			// TODO: turn this into an action/state pair so synchronized
+		}).publish();
+
+		// Then run Epics
+		Set<Observable<RadixNodeAction>> updates = epics.stream()
+			.map(epic -> epic.epic(reducedNodeActions, networkState))
+			.collect(Collectors.toSet());
+		Observable.merge(updates).subscribe(this::dispatch);
```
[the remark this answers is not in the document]
talekhinezh: I assume you mean "Why aren't". Will add a node to implement this later on. Having full support for resource cleanup in the network layer would still require quite a bit of work.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkEpic.java
```diff
@@ -0,0 +1,22 @@
...
+ * Inspired by epics in redux-observables. This follows a similar pattern which can be understood as a creator of a
+ * stream of actions from a stream of actions.
+ *
+ */
+public interface RadixNetworkEpic {
+
+	/**
+	 * Creates a stream of actions from a stream of actions and current/future states.
+	 *
+	 * Note that this should NEVER let an action "slip through" as this will cause an infinite loop.
```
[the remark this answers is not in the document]
talekhinezh: "slip through" as in -emit action which the epic has received-. Updated comments.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -0,0 +1,143 @@
...
+			return this;
+		}
+
+		public RadixNetworkControllerBuilder addEpic(RadixNetworkEpic epic) {
+			this.epics.add(epic);
+			return this;
+		}
+
+		public RadixNetworkController build() {
+			Objects.requireNonNull(network);
```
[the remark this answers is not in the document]
talekhinezh: I think I'd rather have it build a default network rather than requiring it upfront. Will fix.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -0,0 +1,143 @@
...
+	 */
+	@Override
+	public Observable<SubmitAtomAction> submitAtom(Atom atom) {
+		SubmitAtomAction initialAction = SubmitAtomRequestAction.newRequest(atom);
+
+		Observable<SubmitAtomAction> status =
+			nodeActions.ofType(SubmitAtomAction.class)
+				.filter(u -> u.getUuid().equals(initialAction.getUuid()))
+				.takeUntil(u -> u instanceof SubmitAtomResultAction);
+		ConnectableObservable<SubmitAtomAction> replay = status.replay();
```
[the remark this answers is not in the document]
talekhinezh: This is an artifact from the earlier api where it was weird to have a `submitAtom` method which didn't do anything until one called `subscribe` to the result of the method. The replay was added so that this observable can connect immediately with the results being saved for a subscriber to view without losing information.

Having said that, this will be refactored out later on now that there is a generic `dispatch` and `getActions` method in the controller.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/DiscoverMoreNodesAction.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * A dispatchable action which requests for more nodes to be discovered
+ */
+public final class DiscoverMoreNodesAction implements RadixNodeAction {
+	private DiscoverMoreNodesAction() {
+	}
+
+	public static DiscoverMoreNodesAction instance() {
+		return new DiscoverMoreNodesAction();
```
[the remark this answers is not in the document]
talekhinezh: I've thought about creating a singleton but decided to have new instances for now as how this exact action will look like is still unclear.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/DiscoverMoreNodesAction.java
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	// TODO: remove this
+	@Override
+	public RadixNode getNode() {
```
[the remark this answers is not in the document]
talekhinezh: Yup, definitely some issues here which I think can be resolved by a more generic `RadixAction` interface in a future change.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/GetLivePeersResultAction.java
```diff
@@ -0,0 +1,39 @@
...
+public class GetLivePeersResultAction implements JsonRpcResultAction<List<NodeRunnerData>> {
+	private final RadixNode node;
+	private final List<NodeRunnerData> data;
+
+	private GetLivePeersResultAction(RadixNode node, List<NodeRunnerData> data) {
+		Objects.requireNonNull(node);
+		Objects.requireNonNull(data);
+
+		this.node = node;
+		this.data = data;
```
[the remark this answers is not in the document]
talekhinezh: yup, good catch.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/SubmitAtomResultAction.java
```diff
@@ -0,0 +1,92 @@
...
+import com.radixdlt.client.core.network.jsonrpc.RadixJsonRpcClient.NodeAtomSubmissionUpdate;
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * A dispatchable event action which signifies the end result of an atom submission flow
+ */
+public class SubmitAtomResultAction implements SubmitAtomAction {
+	public enum SubmitAtomResultActionType {
+		FAILED(NodeAtomSubmissionState.FAILED),
```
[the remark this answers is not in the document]
talekhinezh: Yes, I like this proposal (and I actually do like a good ole switch statement :) though I don't think there are compiler warnings if there are missing case statements since the compiler doesn't know if there may be additional enums during runtime: https://stackoverflow.com/questions/39225553/complete-switch-on-enum-in-java-produces-missing-return-statement-error

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/RadixJsonRpcAutoCloseEpic.java
```diff
@@ -0,0 +1,34 @@
...
+import com.radixdlt.client.core.network.epics.WebSocketsEpic.WebSockets;
+import io.reactivex.Observable;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Epic which attempts to close a websocket when a Json Rpc method is finished executing.
+ * Note that the websocket won't close if there are still listeners.
+ */
+public class RadixJsonRpcAutoCloseEpic implements RadixNetworkEpic {
+	private final static int DELAY_CLOSE_SECS = 5;
```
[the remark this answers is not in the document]
talekhinezh: I think it's still too early to decide on what these network constants are as I still see a lot of improvement that can be done how to auto close and auto connect, perhaps even without some sort of delay which is why I kept the constants hidden outside of the class.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/RadixJsonRpcMethodsEpic.java
```diff
@@ -0,0 +1,54 @@
...
+									.toMaybe();
+							}
+
+							if (a instanceof GetNodeDataRequestAction) {
+								return jsonRpcClient.getInfo()
+									.map(d -> GetNodeDataResultAction.of(a.getNode(), d))
+									.toMaybe();
+							}
+
+							return Maybe.<RadixNodeAction>empty();
```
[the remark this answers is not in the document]
talekhinezh: No, it just means this epic doesn't support whatever action method was emitted. Should probably turn this class into a `RadixJsonRpcMethodEpic` interface down the line where each instance supports a different method.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/WebSocketsEpic.java
```diff
@@ -0,0 +1,78 @@
...
+	public static class WebSocketsEpicBuilder {
+		private final List<Function<WebSockets, RadixNetworkEpic>> webSocketEpics = new ArrayList<>();
+
+		public WebSocketsEpicBuilder add(Function<WebSockets, RadixNetworkEpic> webSocketEpic) {
+			webSocketEpics.add(webSocketEpic);
+			return this;
+		}
+
+		public WebSocketsEpic build() {
+			return new WebSocketsEpic(new ArrayList<>(webSocketEpics));
```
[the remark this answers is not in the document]
talekhinezh: Just to protect from some messiness if `add()` and `build()` calls are interleaved.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/selector/ConnectionAliveFilter.java
```diff
@@ -0,0 +1,14 @@
+package com.radixdlt.client.core.network.selector;
+
+import com.radixdlt.client.core.network.websocket.WebSocketStatus;
+import com.radixdlt.client.core.network.RadixNodeState;
+
+/**
+ * A connection status filter that filters out inactive peers
+ */
+public class ConnectionAliveFilter implements RadixPeerFilter {
```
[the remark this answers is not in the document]
talekhinezh: Yup true, all of the classes under `selector` package are artifacts from earlier and just there for reminder that these need to be re-implemented down the line.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/reducers/RadixNetwork.java
```diff
@@ -0,0 +1,53 @@
...
+			final RadixNode node = wsEvent.getNode();
+			newMap = new HashMap<>(state.getNodes());
+			newMap.merge(
+				node,
+				RadixNodeState.of(node, WebSocketStatus.valueOf(wsEvent.getType().name())),
+				(old, val) -> RadixNodeState.of(old.getNode(), val.getStatus(), old.getData().orElse(null))
+			);
+		}
+
+		return newMap != null ? new RadixNetworkState(newMap) : null;
```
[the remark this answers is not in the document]
talekhinezh: no, just laziness due to my part where `null` from reducer signifies that networkState has not changed but now realize this can just be done with a `==` check rather than a null check in `RadixNetworkController`.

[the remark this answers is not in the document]
talekhinezh: Will fix.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
talekhinezh: They wouldn't be able to do that since this is exposing an `Observable` rather than a `Subject`

radixdlt-java/src/main/java/com/radixdlt/client/core/network/websocket/WebSocketClient.java
talekhinezh: In practice I haven't seen `onClosing` being called but the state change will occur in the `onClose` callback.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -0,0 +1,143 @@
...
+	private static final Logger LOGGER = LoggerFactory.getLogger(RadixNetworkController.class);
+
+	public static class RadixNetworkControllerBuilder {
+		private RadixNetwork network;
+		private List<RadixNetworkEpic> epics = new ArrayList<>();
+
+		public RadixNetworkControllerBuilder() {
+		}
+
+		public RadixNetworkControllerBuilder network(RadixNetwork network) {
```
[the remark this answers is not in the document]
talekhinezh: Makes sense.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNode.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (!(o instanceof RadixNode)) {
+			return false;
+		}
+
+		RadixNode p = (RadixNode) o;
+		return p.url.equals(this.url);
```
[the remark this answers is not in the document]
talekhinezh: Okay, maybe the description is not entirely accurate as a "Unique" Radix node *really* should mean a node associated with a key so this class should really be `RadixNodeEndpoint` which represents a unique URL (aka it could be the same machine).

Down the line though, this needs to be replaced with the node key and it's known associated endpoints. I don't think we're there yet though.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/SubmitAtomResultAction.java
```diff
@@ -0,0 +1,92 @@
...
+		Objects.requireNonNull(uuid);
+		Objects.requireNonNull(atom);
+		Objects.requireNonNull(node);
+		Objects.requireNonNull(type);
+
+		this.uuid = uuid;
+		this.atom = atom;
+		this.node = node;
+		this.type = type;
+		this.data = data;
```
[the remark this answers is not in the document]
talekhinezh: cool

radixdlt-java/src/main/java/com/radixdlt/client/core/network/jsonrpc/RadixJsonRpcClient.java
flotothemoon: Makes sense

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/RadixJsonRpcMethodsEpic.java
```diff
@@ -0,0 +1,54 @@
...
+					final WebSocketClient ws = webSockets.get(a.getNode());
+					return ws.getState()
+						.filter(WebSocketStatus.CONNECTED::equals)
+						.firstOrError()
+						.flatMapMaybe(i -> {
+							RadixJsonRpcClient jsonRpcClient = new RadixJsonRpcClient(ws);
+							if (a instanceof GetLivePeersRequestAction) {
+								return jsonRpcClient.getLivePeers()
+									.map(l -> GetLivePeersResultAction.of(a.getNode(), l))
+									.toMaybe();
```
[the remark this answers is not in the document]
talekhinezh: refactored and cleaned up :)

talekhinezh:
@laszlourszuly Re: JavaDoc compliance, yes, thanks for the reminder! I think this will be a work in progress for me and any reminder specific comments are appreciated. Regarding use of 2nd person, I'll fix the ones I see.

PR #134, opened 2019-01-10
Adding .java-version file which jenv will read and set the Java versi…

Sajjon:
Adding .java-version file which [jenv](http://www.jenv.be/) will read and set the Java version when ever you navigate the Radix Java directory.


flotothemoon:
While I don't see any issue with approving this particular PR, I'm worried about the precedent. Which 3rd party tools do we want to cater to? How are they selected? Inevitable, someone in the community will feel left out because we directly support X but not their favorite Y.

Sajjon:
@flotothemoon Sure are always flame wars going on between devs. But as for Java version management I do not think there is any alternative to Jenv for macOS? 

I was just hoping this would make things a little bit easier for us. Giving the ability to control the Java versions used by us devs in the source code itself.

However, we also have a Java version set in the [build.gradle file](https://github.com/radixdlt/radixdlt-java/blob/develop/build.gradle#L30-L31)

```
sourceCompatibility = 1.8
targetCompatibility = 1.8
```
But I believe that to be a more project build config, where as the `.java-version` file helps e.g. IntelliJ picking up the correct JDK path.

laszlourszuly:
@Sajjon @flotothemoon Is this something that should be merged?

Sajjon:
@laszlourszuly I think it is always a good idea to enforce certain versions of a language in Git. If you don't have [jenv](http://www.jenv.be/), just install it and then your computer will load the correct Java version as soon as you CD into the project directory. What's not to like? :)

laszlourszuly:
Makes sense. Merging as it's a non-intrusive, easy to revert change (would any unexpected issues arise).

PR #135, opened 2019-01-15
Feature/particle groups

flotothemoon:
Introduction of ParticleGroup: an explicit grouping of related Particles within an Atom. ParticleGroups reside between Atoms and Particles in hierarchy. An atom can only store ParticleGroups, all Particles must therefore be in a ParticleGroup. This will allow us to process and validate larger Atoms with many transactions more effectively while also providing better UX by giving explicit intent and relations of particles in an Atom.


flotothemoon:
> 🍪🍪🍪 for the formal documentation on new, public methods. I do think, however, that there is a value in conforming to standard patterns and grammar notations, for example:
> 
> > **Use 3rd person (descriptive) not 2nd person (prescriptive).**
> > The description is in 3rd person declarative rather than 2nd person imperative.
> > Gets the label. (preferred)
> > Get the label. (avoid)
> 
> or, where applicable:
> 
> > **First Sentence**
> > The first sentence of each doc comment should be a summary sentence, containing a concise but complete description of the API item. This means the first sentence of each member, class, interface or package description. The Javadoc tool copies this first sentence to the appropriate member, class/interface or package summary. This makes it important to write crisp and informative initial sentences that can stand on their own.
> > This sentence ends at the first period that is followed by a blank, tab, or line terminator, or at the first tag [...]
> 
> Reference: https://www.oracle.com/technetwork/java/javase/index-137868.html
> 
> I realise that we probably should have a discussion on documentation, as well as coding style guides at some point (more about which standard to pick, rather than which exotic combination of standards to come up with). I also realise that this discussion won't happen anytime before public launch. But it would be nice if we could start thinking about it a bit...
> 
> Best of regards,
> -- The JavaDoc nazi

I sincerely appreciate your efforts @laszlourszuly and I agree with all your points. I make an effort to include documentation for all public-facing endpoints because I agree it's important. I also like consistency a lot, so I too think we should agree on a standard for all things formatting and documentation. I however also think that it's not a priority right now and will become more important the closer we get to public launch. Thanks for your input :)

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -0,0 +1,141 @@
...
+			Objects.requireNonNull(particle, "particle is required");
+			Objects.requireNonNull(spin, "spin is required");
+
+			SpunParticle<?> spunParticle = SpunParticle.of(particle, spin);
+			this.particles.add(spunParticle);
+
+			return this;
+		}
+
+		public final boolean removeParticle(SpunParticle<?> spunParticle) {
```
talekhinezh: I don't think we want to expose ability to remove state from a builder as it would encourage bad uses and it doesn't look like this method is being used so I'd prefer to just remove it.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -0,0 +1,141 @@
...
+			Objects.requireNonNull(spunParticle, "spunParticle is required");
+
+			return this.particles.remove(spunParticle);
+		}
+
+		public ParticleGroup build() {
+			return new ParticleGroup(ImmutableList.copyOf(this.particles));
+		}
+
+		public boolean hasParticles() {
```
talekhinezh: Not sure about use of this method either. Builders should be simple and logic about construction should be done outside of builder.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/StatefulActionToParticleGroupsMapper.java
```diff
@@ -58,7 +59,7 @@ public RadixAddress address() {
 	 * actions to be included in the current transaction.
 	 *
 	 * @param action the current action
-	 * @param store application state requested as specified by requiredState()
+	 * @param store  application state requested as specified by requiredState()
```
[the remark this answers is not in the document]
flotothemoon: Space is cool, I guess. Removed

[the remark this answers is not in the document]
flotothemoon: Reason is to align the parameter description vertically.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * A group of particles representing one intent, e.g. a transfer.
+ * <p>
+ * * @author flotothemoon
+ */
+@SerializerId2("PARTICLEGROUP")
+public class ParticleGroup extends SerializableObject {
+	@JsonProperty("particles")
+	@DsonOutput(DsonOutput.Output.ALL)
+	private final List<SpunParticle> particles = new ArrayList<>();
```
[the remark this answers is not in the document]
flotothemoon: I thought it was about time to discontinue that. The space savings are minimal and introduce an unnecessary edge case for no measurable benefit.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
flotothemoon: Good catch, forgot to remove it after copying the code over from ParticleGroup when introducing builders.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
flotothemoon: Good point, removed in Core and Library.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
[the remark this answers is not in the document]
talekhinezh: I think decisions like this need to be discussed with all and should be implemented as a separate task as there can always be potential unintended side effects.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -0,0 +1,141 @@
...
+
+	private ParticleGroup() {
+	}
+
+	public ParticleGroup(List<SpunParticle> particles) {
+		Objects.requireNonNull(particles, "particles is required");
+
+		this.particles.addAll(particles);
```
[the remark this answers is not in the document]
flotothemoon: Made immutable

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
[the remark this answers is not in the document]
flotothemoon: Good point, do you think I should remove it and reimplement that change in a separate PR? Or just leave it in?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -257,15 +262,15 @@ public RadixApplicationAPI build() {
...
-				statelessActionToParticlesMappers,
-				statefulActionToParticlesMappers,
-				reducers,
-				atomMappers,
-				atomErrorMappers
+					identity,
+					universe,
+					feeMapper,
+					ledger,
+					this.statelessActionToParticleGroupsMappers,
```
mpsandiford: So much diff, so little change!

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -324,7 +338,7 @@ public static RadixApplicationAPI create(RadixIdentity identity) {
 	 * @return Disposable to dispose to stop pulling
 	 */
 	public Disposable pull() {
-		return pull(getMyAddress());
+		return this.pull(this.getMyAddress());
```
mpsandiford: These changes are really unhelpful.  Is this another save action thing that needs to be turned off?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -368,26 +382,26 @@ public TokenClassReference getNativeTokenRef() {
 	 * address automatically until the observable is disposed.
 	 *
 	 * @param actionClass the Action class
-	 * @param address the address to retrieve the state of
-	 * @param <T> the Action class
+	 * @param address     the address to retrieve the state of
+	 * @param <T>         the Action class
 	 * @return a hot observable of the actions at the given address
 	 */
 	public <T> Observable<T> getActions(Class<T> actionClass, RadixAddress address) {
```
mpsandiford: There doesn't appear to be any meaningful changes in this method.  Why so much diff?

This really needs to be sorted out.  It's a huge waste of review time.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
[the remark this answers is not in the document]
flotothemoon: Changed to immutableList and leaving in

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
flotothemoon: :c

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
flotothemoon: I've disabled everything there is to disable on formatting actions, save actions, indent actions. I hope this will stop these annoyances. I think this is just to align the argument descriptions, but I agree, unnecessary.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
flotothemoon: As said in another comment, I've disabled everything there is to disable in terms of reformatting and auto-actions. I hope this fixes this issue. Sorry about that.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/unique/AlreadyUsedUniqueIdReasonMapper.java
```diff
@@ -28,4 +34,30 @@
...
+			String particleIndexStr = pointerToIssue.substring(pointerToIssue.lastIndexOf('/') + 1);
+
+			int groupIndex = Integer.parseInt(groupIndexStr);
+			int particleIndex = Integer.parseInt(particleIndexStr);
+
+			ParticleGroup particleGroup = atom.particleGroups().collect(Collectors.toList()).get(groupIndex);
+
+			return Optional.of((SpunParticle<?>) particleGroup.spunParticles().collect(Collectors.toList()).get(particleIndex));
+		} catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
+			LoggerFactory.getLogger(AlreadyUsedUniqueIdReasonMapper.class).error("Malformed pointerToIssue");
```
mpsandiford: Would prefer to see this done in the usual way, static field with logger initialised at class load.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * A group of particles representing one intent, e.g. a transfer.
+ * <p>
+ * * @author flotothemoon
+ */
+@SerializerId2("PARTICLEGROUP")
+public class ParticleGroup extends SerializableObject {
+	@JsonProperty("particles")
+	@DsonOutput(DsonOutput.Output.ALL)
+	private final ImmutableList<SpunParticle> particles;
```
mpsandiford: Noice.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -0,0 +1,119 @@
...
+
+	private ParticleGroup() {
+		this.particles = ImmutableList.of();
+	}
+
+	public ParticleGroup(List<SpunParticle> particles) {
```
mpsandiford: This could be a Collection or even Iterable.

flotothemoon: I thought I did that already. Huh.

PR #137, opened 2019-01-21
Test signing capabilites of seeded key pairs

laszlourszuly:
Adding test as per sprint demo feedback.


laszlourszuly:
Thanks for the quick review, @talekhinezh!

PR #138, opened 2019-01-23
Adding expected expect hexstring for hash for empty model and serialised model.

Sajjon:
Adding comparisons in tests of atoms for hash, against an expected hex string, so that we more easily can verify correctness between libraries.

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
@@ -4,13 +4,13 @@ import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
-import okio.ByteString;
 import org.bouncycastle.util.encoders.Base64;
 import org.radix.common.ID.EUID;
 
 import com.radixdlt.client.core.crypto.ECPublicKey;
 import com.radixdlt.client.core.crypto.ECSignature;
 import com.radixdlt.client.core.util.Hash;
+import org.radix.utils.primitives.Bytes;
 
 public final class RadixHash {
 	private final byte[] hash;
@@ -75,5 +75,7 @@ public final class RadixHash {
 		return Base64.toBase64String(hash);
 	}
 
-	public String toHexString() { return ByteString.of(hash).hex(); }
+	public String toHexString() {
+		return Bytes.toHexString(hash);
+	}
 }
--- a/radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
@@ -4,6 +4,7 @@ import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
+import okio.ByteString;
 import org.bouncycastle.util.encoders.Base64;
 import org.radix.common.ID.EUID;
 
@@ -73,4 +74,6 @@ public final class RadixHash {
 	public String toString() {
 		return Base64.toBase64String(hash);
 	}
+
+	public String toHexString() { return ByteString.of(hash).hex(); }
 }
--- a/radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/AtomTest.java
+++ b/radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/AtomTest.java
@@ -19,6 +19,8 @@ public class AtomTest {
 		assertTrue(atom.getDataParticles().isEmpty());
 		assertTrue(atom.getConsumables(Spin.UP).isEmpty());
 		assertTrue(atom.getConsumables(Spin.DOWN).isEmpty());
+		/// The origin of these hashes are this library it self, commit: acbc5307cf5c9f7e1c30300f7438ef5dbc3bb629
+		/// These hashes can be used as a reference for other Radix libraries, e.g. Swift.
 		assertEquals("bf6a73657269616c697a65721a001ed1516776657273696f6e1864ff", ByteString.of(atom.toDson()).hex());
 		assertEquals("1b1cff72cb4f79d2eb50b5fb2777d65bebb5cad146e2006f25cde7a53445ffe7", atom.getHash().toHexString());
 		assertEquals("1b1cff72cb4f79d2eb50b5fb2777d65b", atom.getHid().toHexString());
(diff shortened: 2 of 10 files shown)


radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/AtomTest.java
```diff
@@ -16,6 +19,7 @@ public void testEmptyAtom() {
 		assertTrue(atom.getDataParticles().isEmpty());
 		assertTrue(atom.getConsumables(Spin.UP).isEmpty());
 		assertTrue(atom.getConsumables(Spin.DOWN).isEmpty());
+		assertEquals("bf6a73657269616c697a65721a001ed1516776657273696f6e1864ff", ByteString.of(atom.toDson()).hex());
```
laszlourszuly: Maybe this assertion would benefit from a short comment where this hex-hash comes from. Is it generated by the JS-library? Kotlin? Some external tool?

radixdlt-java/src/test/java/com/radixdlt/client/util/HashTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static junit.framework.TestCase.assertEquals;
+
+public class HashTest {
+
+    @Test
+    public void testDoubleSha256Hash() {
+        byte[] data = new byte[0];
+        try {
+            data = "Hello Radix".getBytes("UTF-8");
```
laszlourszuly: You could use `StandardCharsets.UTF_8` instead of the string literal to avoid the try-catch block. Because readability.

Furthermore, if you prefer to keep the string literal, I'd much rather let the test case fail, (basically adding a general `throws Exception` clause to the method signature and remove the try + catch blocks).

radixdlt-java/src/test/java/com/radixdlt/client/util/HashTest.java
```diff
@@ -0,0 +1,26 @@
...
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+        }
+        byte[] singleHash = Hash.sha256(data);
+        byte[] doubleHash = Hash.sha256(singleHash);
+        assertEquals("374d9dc94c1252acf828cdfb94946cf808cb112aa9760a2e6216c14b4891f934", ByteString.of(singleHash).hex());
+        assertEquals("fd6be8b4b12276857ac1b63594bf38c01327bd6e8ae0eb4b0c6e253563cc8cc7", ByteString.of(doubleHash).hex());
+    }
```
laszlourszuly: Maybe consider adding a similar "origin of hash" comment here as well. It will come in handy when researching why this test failed (once it fails unexpectedly).

radixdlt-java/src/test/java/com/radixdlt/client/util/HashTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.Test;
+
+import java.io.UnsupportedEncodingException;
+
+import static junit.framework.TestCase.assertEquals;
+
+public class HashTest {
+
+    @Test
+    public void testDoubleSha256Hash() {
```
laszlourszuly: Generally I think it's a good idea to name your unit test cases on the form:

```java
@Test
public void if_i_do_this__then_that_should_happen() {
    // Test something
}
```

I think this helps both you, as a test writer, to stay on focus, but even more so me, as a reviewer or peer developer, to understand what you intend to test and to verify that you're really testing what you're claiming.

This is a matter of taste though, and to be honest I'd value consistency (with other tests in the suite) over clarity in this case.

laszlourszuly:
Sorry for barging in, uninvited. I had a couple of minutes over 'till nest task starts... 😄

radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/AtomTest.java
Sajjon: @laszlourszuly sorry missed your review comments :) 

This is, currently, the value generated by this library. I.e. if we make changes in the Atom, it will result in new Hash and thus the test will need to get updated.

radixdlt-java/src/test/java/com/radixdlt/client/util/HashTest.java
Sajjon: Actually I tried that :D but I failed to import the package, must have been a type, got it working now :) 

Using the StandardCharset results in the `getBytes` not throwing, which is great.

Sajjon:
@laszlourszuly Have another look! I think I've fixed everything :)

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
```diff
@@ -4,6 +4,7 @@
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
+import okio.ByteString;
```
mpsandiford: Not really happy with this dependency being here.  Can we use `Bytes.toHexString(...)`?

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
```diff
@@ -73,4 +74,6 @@ public int hashCode() {
 	public String toString() {
 		return Base64.toBase64String(hash);
 	}
+
+	public String toHexString() { return ByteString.of(hash).hex(); }
```
mpsandiford: Would prefer to see this following the coding conventions:
```
public String toHexString() {
    return Bytes.toHexString(hash);
}
````

File: radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/AtomTest.java [hunk 7]
```diff
@@ -5,8 +5,11 @@
 import static org.junit.Assert.assertTrue;
 
 import com.radixdlt.client.core.atoms.particles.Spin;
+import com.radixdlt.client.core.util.Hash;
+import okio.ByteString;
```
mpsandiford: Again, seems like an unnecessary additional dependency on an external library.  Can we use Bytes.toHexString(...) instead?

radixdlt-java/src/test/java/com/radixdlt/client/util/HashTest.java
```diff
@@ -0,0 +1,25 @@
+package com.radixdlt.client.util;
+
+import com.radixdlt.client.core.util.Hash;
+import okio.ByteString;
```
mpsandiford: Please use Bytes.toHexString(...).

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/RadixHash.java
Sajjon: Absolutely, I got confused by Intellij which collapsed e.g. `hashCode` into one line. Got it!

Sajjon:
@mpsandiford I've removed all references to okio

PR #139, opened 2019-01-24
Renaming method in Atom having obsolete names such as "consumables" (…

Sajjon:
…instead of the actual "OwnedTokensParticle") and renamed "getDataParticle" to "getMessageParticles"


talekhinezh:
@Sajjon Looks good to me. Merging...

PR #143, opened 2019-02-08
Refactor AccountableQuark into an interface

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/atommodel/message/MessageParticle.java
```diff
@@ -72,15 +78,21 @@ private MessageParticle() {
...
 		this.from = Objects.requireNonNull(from, "from is required");
+		this.to = Objects.requireNonNull(to, "to is required");
 		this.bytes = Arrays.copyOf(bytes, bytes.length);
 		this.metaData.putAll(metaData);
 	}
 
+	@Override
+	public Set<RadixAddress> getAddresses() {
+		return ImmutableSet.of(from, to);
+	}
```
laszlourszuly: I'm lacking the context maybe, but how would I use the output of this method? I think `Set`'s don't provide any guarantees for the order of items. How would I, for example, distinguish one address from the other in the resulting set?

talekhinezh: This interface is just supposed to expose which addresses this particle should be be stored in. The `getTo` and `getFrom` methods in this class should be what you use to retrieve the specific addresses.

PR #144, opened 2019-02-11
Add metaData to atoms and particle groups

MuncleUscles:


radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -43,6 +43,10 @@
 	@DsonOutput(value = {DsonOutput.Output.API, DsonOutput.Output.WIRE, DsonOutput.Output.PERSIST})
 	private final Map<String, ECSignature> signatures = new HashMap<>();
 
+	@JsonProperty("metaData")
+	@DsonOutput(DsonOutput.Output.ALL)
+	private Map<String, String> metaData = new HashMap<>();
```
mpsandiford: Could be final.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -9,9 +9,7 @@
 import org.radix.serialization2.SerializerId2;
 import org.radix.serialization2.client.SerializableObject;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
mpsandiford: Would prefer if wildcard imports were not used.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -202,4 +216,8 @@ public int hashCode() {
 	public String toString() {
 		return "Atom (" + this.getHid().toString() + ")";
 	}
+
```
mpsandiford: In an ideal world, this would be closer to the other getters.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -112,8 +134,18 @@ public final ParticleGroupBuilder addParticle(Particle particle, Spin spin) {
...
+
+			this.metaData.put(key, value);
+
+			return this;
+		}
+
+
 		public ParticleGroup build() {
-			return new ParticleGroup(ImmutableList.copyOf(this.particles));
+			return new ParticleGroup(ImmutableList.copyOf(this.particles), metaData);
```
mpsandiford: Appreciate this is old code, but I'm not sure we need to copy the list here and again in the constructor.

radixdlt-java/src/test/java/com/radixdlt/client/core/atoms/MessageParticleTest.java
```diff
@@ -16,6 +16,7 @@ public void testApplicationMetaData() {
 		MessageParticle messageParticle = new MessageParticleBuilder()
 			.payload(new byte[0])
 			.from(mock(RadixAddress.class))
+			.to(mock(RadixAddress.class))
```
mpsandiford: Why did this change?

MuncleUscles: Because the unit test was broken when I tried to run them

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
MuncleUscles: I noticed that, I should have fixed it right away

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -202,4 +216,8 @@ public int hashCode() {
...
+	public Map<String, String> getMetaData() {
+		return this.metaData;
```
laszlourszuly: I've noticed that we sometimes prefer to return unmodifiable copies of `List`'s and `Set`'s. I'm not sure if that pattern would also apply to `Map`'s as well, or if it's relevant in this case at all (do we expose this to user facing abstractions, for example?)

What's your take @mpsandiford?

MuncleUscles: I think it would make a lot of sense here, what's the best way to do that?

mpsandiford: `ImmutableMap` for the declaration of `this.metaData` would be best.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -183,6 +200,12 @@ private static BigInteger ownedTokensToBigInteger(SpunParticle<OwnedTokensPartic
 		return (value.getSpin() == Spin.UP) ? bi : bi.negate();
 	}
 
+
+
+	public ImmutableMap<String, String> getMetaData() {
```
mpsandiford: As noted elsewhere, I would like to see this as a `Map<String, String>` so that Guava types are not leaked into the public interface.  I think this issue is more important to resolve here in the library.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/ParticleGroup.java
```diff
@@ -52,13 +74,30 @@ public ParticleGroup(Iterable<SpunParticle> particles) {
 		return this.spunParticles().filter(p -> p.getSpin() == spin).map(SpunParticle::getParticle);
 	}
 
+	/** Get the metadata associated with the particle group
+	 * @return a map of the metadata
+	 */
+	public ImmutableMap<String, String> getMetaData() {
```
mpsandiford: Would like to see this as a `Map<String, String>` as noted elsewhere.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -183,6 +200,12 @@ private static BigInteger ownedTokensToBigInteger(SpunParticle<OwnedTokensPartic
 		return (value.getSpin() == Spin.UP) ? bi : bi.negate();
 	}
 
+
+
```
mpsandiford: Muchos vertical whitespace.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/PowFeeMapper.java
```diff
@@ -24,21 +25,21 @@
 public class PowFeeMapper implements FeeMapper {
 	private static final int LEADING = 16;
 
-	private final Function<List<ParticleGroup>, RadixHash> hasher;
+	private final Function<Atom, RadixHash> hasher;
 	private final ProofOfWorkBuilder powBuilder;
 
-	public PowFeeMapper(Function<List<ParticleGroup>, RadixHash> hasher, ProofOfWorkBuilder powBuilder) {
+	public PowFeeMapper(Function<Atom, RadixHash> hasher, ProofOfWorkBuilder powBuilder) {
```
mpsandiford: This interface change needs some changes in PowFeeMapperTest that haven't been made, and prevent the unit tests from compiling / running at the moment.  Probably a good habit to get into running `gradle check` before pushing.

PR #145, opened 2019-02-19
Move timestamp to Atom metadata, remove TimestampParticle

flotothemoon:
This removes the TimestampParticle and introduces a special "timestamp" field containing a single 64-bit integer.
See RLAU-576


radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -700,22 +700,21 @@ public Result transferTokens(
...
 			.<List<ParticleGroup>>scanWith(
 					ArrayList::new,
 					(a, b) -> Stream.concat(a.stream(), Stream.of(b)).collect(Collectors.toList())
 			)
 			.lastOrError()
 			.map(particleGroups -> {
 				List<ParticleGroup> allParticleGroups = new ArrayList<>(particleGroups);
-				Atom atom = new Atom(particleGroups);
+				ImmutableMap<String, String> metaData = ImmutableMap.of(
+					Atom.METADATA_TIMESTAMP_KEY, String.valueOf(System.currentTimeMillis()));
```
mpsandiford: At some point in the future we will want to be using RTP time here, so it would be good to abstract this out into another method, perhaps in a new utility class or even an injectable `LongSupplier`.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -49,14 +50,19 @@
 	private final ImmutableMap<String, String> metaData;
 
 	private Atom() {
-		this.metaData = ImmutableMap.of();
+		this(0);
```
mpsandiford: Would it be better to use an empty map here, rather than going to the unnecessary trouble of creating a new map with a timestamp that is invalid?

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -49,14 +50,19 @@
...
+	}
+
+	public Atom(long timestamp) {
+		this.metaData = ImmutableMap.of(METADATA_TIMESTAMP_KEY, String.valueOf(timestamp));
 	}
 
 	public Atom(List<ParticleGroup> particleGroups) {
```
mpsandiford: It's not possible to create an atom that will pass validation with this constructor.  I'm not sure it's necessary or useful.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomValidator.java
```diff
@@ -69,8 +69,12 @@ public void validateSignatures(Atom atom) throws AtomValidationException {
...
 	public void validate(Atom atom) throws AtomValidationException {
 		// TODO: check with universe genesis timestamp
-		if (atom.getTimestamp() == null || atom.getTimestamp() == 0L) {
-			throw new AtomValidationException("Null or Zero Timestamp");
+		if (!atom.hasTimestamp()) {
+			throw new AtomValidationException("Atom does not contain a timestamp");
+		}
+
+		if (atom.getTimestamp() == 0L) {
+			throw new AtomValidationException("Atom timestamp cannot be zero");
```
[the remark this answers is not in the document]
flotothemoon: Sure, I just translated it from what was there before. Will change.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
flotothemoon: Fixed, good catch

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
flotothemoon: Cool, will do.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
flotothemoon: Changed as proposed by Laszlo, good catch

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomValidator.java
```diff
@@ -69,8 +69,12 @@ public void validateSignatures(Atom atom) throws AtomValidationException {
 	@Override
 	public void validate(Atom atom) throws AtomValidationException {
 		// TODO: check with universe genesis timestamp
-		if (atom.getTimestamp() == null || atom.getTimestamp() == 0L) {
-			throw new AtomValidationException("Null or Zero Timestamp");
+		if (!atom.hasTimestamp()) {
+			throw new AtomValidationException("Atom does not contain a timestamp");
+		}
+
+		if (atom.getTimestamp() < 0L) {
```
mpsandiford: Either the comparator should be `<=`, or the error message should change here.  I think the `<=` is more consistent with previous checks.

PR #147, opened 2019-02-22
Feature/rlau 651 refactor fungible to interface

flotothemoon:


radixdlt-java/src/main/java/com/radixdlt/client/atommodel/Fungible.java
```diff
@@ -0,0 +1,16 @@
+package com.radixdlt.client.atommodel;
+
+import org.radix.utils.UInt256;
+
+/**
+ * Temporary interface for representing an amount of interchangeable Particles.
```
mpsandiford: Why is this temporary?  What conditions are there for it getting removed?

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/Fungible.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+public interface Fungible {
+	UInt256 getAmount();
```
mpsandiford: Should really have some Javadoc for each of these.

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/Fungible.java
flotothemoon: Good question. All the other interfaces quark interfaces are marked temporary as well, so I assume @talekhinezh has some plan to change them in the near future.

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/Fungible.java
flotothemoon: Usually, I didn't include any in this case because I assumed it would be changed in the next version anyway.

PR #153, opened 2019-03-03
Feature/rlau 981 split owned tokens

flotothemoon:
This is the first step in the fungible refactor plan. The overall steps are

 1. **Split OwnedTokensParticle into explicit Minted, Transferred, BurnedTokensParticles**
 2. Remove FungibleType in favour of the explicit particle types
 3. Introduce a fungible system call into the constraint machine
 4. Transfer fungible logic of the tokens to that system call
 5. Transfer all remaining tokens instances constraints into constraint scrypts


flotothemoon:
This is the first step in the fungible refactor plan. The overall steps are

 1. **Split OwnedTokensParticle into explicit Minted, Transferred, BurnedTokensParticles**
 2. Remove FungibleType in favour of the explicit particle types
 3. Introduce a fungible system call into the constraint machine
 4. Transfer fungible logic of the tokens to that system call
 5. Transfer all remaining tokens instances constraints into constraint scrypts

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Hashable.java
```diff
@@ -0,0 +1,23 @@
+package com.radixdlt.client.core.atoms;
+
+import org.radix.common.ID.EUID;
+
+/**
+ * Representing any type that can be *hashed* into a {@link RadixHash}
+ * with a utility method to obtain an HID from that hash.
+ */
+public interface Hashable {
```
talekhinezh: This interface isn't correct.

First issue is that it isn't really being used as an interface but rather for inheritance but now in a worse way because the interface suggests that `hid()` can be modified.

Secondly, the hash of an object shouldn't be calculated by the class itself as we have a very specific mechanism with DSON on how to calculate these hashes and Hash calculation and classes should be decoupled in the long run so these methods shouldn't be supported by an interface.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokenclasses/TokenTypesReducer.java
```diff
@@ -58,23 +54,28 @@ public TokenClassesState reduce(TokenClassesState state, SpunParticle s) {
...
-		} else {
-			OwnedTokensParticle mintedOrBurned = (OwnedTokensParticle) p;
-			BigInteger mintedOrBurnedAmount = UInt256s.toBigInteger(mintedOrBurned.getAmount());
-			BigDecimal change = TokenClassReference.subunitsToUnits(
-				(mintedOrBurned.getType() == FungibleType.BURNED)
-					? mintedOrBurnedAmount.negate()
-					: mintedOrBurnedAmount
-			);
+		} else if (s.getSpin() == Spin.UP) {
+			if (p instanceof MintedTokensParticle || p instanceof BurnedTokensParticle) {
```
talekhinezh: Would rather have all particle class checks occur in the same "level" e.g.:
```
if (p instanceof TokenParticle) {
...
} else if (p instanceof MintedTokensParticle || p instanceof BurnedTokensParticle) {
    if (s.getSpin() == Spin.UP) {
...
    }
}
```

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -122,6 +126,8 @@ public void dispatch(RadixNodeAction action) {
 	public Observable<SubmitAtomAction> submitAtom(Atom atom) {
 		SubmitAtomAction initialAction = SubmitAtomRequestAction.newRequest(atom);
 
+		TestUtils.dumpJsonForHash(atom);
```
talekhinezh: Remove debug code

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/particles/Particle.java
```diff
@@ -40,15 +41,12 @@ protected Particle() {
...
 	}
 
-	public byte[] toDson() {
+	private byte[] toDson() {
 		return Serialize.getInstance().toDson(this, DsonOutput.Output.HASH);
 	}
 
-	public RadixHash getHash() {
+	@Override
+	public RadixHash hash() {
```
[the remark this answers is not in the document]
flotothemoon: Was for the interface, can revert when I remove it again

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Hashable.java
flotothemoon: Fair point. Will remove.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
```diff
@@ -197,20 +201,53 @@ public EUID getHid() {
...
+	private BigInteger consumableToAmount(Pair<ConsumableTokens, Spin> tokens) {
+		BigInteger amount = UInt256s.toBigInteger(tokens.getFirst().getAmount());
+		return tokens.getSecond() == Spin.DOWN ? amount.negate() : amount;
 	}
 
-	private static BigInteger ownedTokensToBigInteger(SpunParticle<OwnedTokensParticle> value) {
-		BigInteger bi = UInt256s.toBigInteger(value.getParticle().getAmount());
-		return (value.getSpin() == Spin.UP) ? bi : bi.negate();
+	private BigInteger consumingToAmount(Pair<ConsumingTokens, Spin> tokens) {
+		BigInteger amount = UInt256s.toBigInteger(tokens.getFirst().getAmount());
```
mpsandiford: I don't think this is correct as it stands now.  There are two types of ConsumingTokens, Burned and Transferred.  Unfortunately they have different signs from each other — UP Transferred tokens mean an increase, whereas UP Burned tokens means a decrease.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/particles/Particle.java
```diff
@@ -40,15 +41,12 @@ protected Particle() {
...
 		return RadixHash.of(toDson());
 	}
-
-	public EUID getHid() {
```
mpsandiford: This really should go back in and be final.

radixdlt-java/src/main/resources/universe/betanet.json
```diff
@@ -148,12 +148,11 @@
                         "particle": {
                             "amount": ":u20:1000000000000000000000000000",
                             "address": ":adr:JHdWTe8zD2BMWwMWZxcKAFx1E8kK3UqBSsqxD9UWkkVD78uMCei",
```
mpsandiford: Using the wrong key.  Should end in Xnknor, not 8uMCei.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokenclasses/TokenTypesReducer.java
flotothemoon: Ha, I specifically didn't nest it out because I like it better that way. But alas, I'm outvoted..

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokenclasses/BurnTokensActionMapper.java
```diff
@@ -78,50 +80,77 @@ private ParticleGroup map(BurnTokensAction burnTokensAction, TokenBalanceState c
...
-			))
+			.map(e -> {
+				if (e.getKey() == null) {
+					return new BurnedTokensParticle(
+						e.getValue(),
+						granularity,
+						burnTokensAction.getTokenTypeReference().getAddress(),
+						System.nanoTime(),
+						burnTokensAction.getTokenTypeReference(),
+						System.currentTimeMillis() / 60000L + 60000L
```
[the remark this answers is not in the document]
flotothemoon: Definitely add an issue, and good point. It's annoying and ugly.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/particles/Particle.java
flotothemoon: Yup, was a bad idea.

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/tokens/ConsumingTokens.java
```diff
@@ -0,0 +1,24 @@
...
+import com.radixdlt.client.core.atoms.Hashable;
+import org.radix.utils.UInt256;
+
+/**
+ * Particle types that *consumes* tokens instances
+ */
+public interface ConsumingTokens extends Fungible, Ownable, Hashable {
+	/**
+	 * Get the token type that this Particle type consumes
+	 * @return THe consumed type
```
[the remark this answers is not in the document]
flotothemoon: I'm impressed you noticed this.

radixdlt-java/src/main/java/com/radixdlt/client/core/atoms/Atom.java
flotothemoon: Fixed, this logic only applies to *consumables* - UP means increase, DOWN means decrease, always. I had my consumables/consuming mixed up.. the irony.

flotothemoon:
Thanks for the feedback @mpsandiford @talekhinezh, concerns have been addressed: Hashable is removed, consumable/consuming logic fixed, formatting / style changes applied.

PR #154, opened 2019-03-04
Feature/rlau 999 remove fungibletype

flotothemoon:
This is the second step in the fungible refactor plan. The overall steps are

1. Split OwnedTokensParticle into explicit Minted, Transferred, BurnedTokensParticles*

2. --> **Remove FungibleType in favour of the explicit particle types**

3. Introduce a fungible system call into the constraint machine

4. Transfer fungible logic of the tokens to that system call

5. Transfer all remaining tokens instances constraints into constraint scrypts


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokenclasses/CreateTokenToParticleGroupsMapper.java
```diff
@@ -4,7 +4,7 @@
 import com.radixdlt.client.application.translate.Action;
 import com.radixdlt.client.application.translate.StatelessActionToParticleGroupsMapper;
 import com.radixdlt.client.application.translate.tokenclasses.CreateTokenAction.TokenSupplyType;
-import com.radixdlt.client.atommodel.FungibleType;
+import com.radixdlt.client.atommodel.tokens.TokensKind;
```
laszlourszuly: Is `TokensKind` an already agreed name? If so please disregard the following.

I think it sounds a bit odd and I'm a bit unsure if "kind" is to be read as "type" or if it implies a higher level of ordering or grouping. Given what I read from how it's used below I would expect something that emphasises `operation` or `action`, but maybe those words are already reserved elsewhere in the domain...

But, again, this is a highly opinionated piece of feedback, on top of that coming from someone who isn't that involved in the core details, so please don't take it too seriously.

flotothemoon: As discussed, I think I will go with a different route and remove TokensKind alltogether and instead only use the class of the Particle. Thanks for the feedback :)

PR #155, opened 2019-03-06
updating second field from travis-ci.org to .com

orbl:
Testing that build triggers on pull request after migration.


mpsandiford:
Merging this as it seems OK.

PR #157, opened 2019-03-11
Remove debouncer and Reducers before Actions

talekhinezh:
* Removed debouncer which makes using system feel more "alive" and faster. For example, integration tests have been reduced from a 5 minute runtime to 4 minute runtime
* Particle reducers are run before action are emitted which fixes a lot of the race conditions seen in client side apps


radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -22,6 +23,11 @@
 	 */
 	private final ConcurrentHashMap<RadixAddress, ReplaySubject<AtomObservation>> cache = new ConcurrentHashMap<>();
 
+	/**
+	 * The In Memory Atom Data Store
```
mpsandiford: This doesn't seem like a very apt comment here.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -42,26 +66,36 @@ public void store(RadixAddress address, AtomObservation atomObservation) {
...
+
 						if (observation.getAtom() != null) {
-							Type curState = curAtomState.get(observation.getAtom().getHash());
-							final boolean shouldUpdate;
-							if (curState == null) {
-								shouldUpdate = observation.getType() == Type.STORE;
+							AtomObservationUpdateType nextUpdate = observation.getUpdateType();
+							AtomObservationUpdateType lastUpdate = atomsObservationState.get(observation.getAtom());
+
+							final boolean filter;
```
mpsandiford: I think the sense of this boolean is back to front.  It should be called `include` or something similar.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -58,14 +61,19 @@ public RadixNetworkController build() {
...
+
 	private RadixNetworkController(RadixNetwork network, List<RadixNetworkEpic> epics) {
 		Objects.requireNonNull(network);
 		Objects.requireNonNull(epics);
 
 		this.networkState = BehaviorSubject.createDefault(new RadixNetworkState(Collections.emptyMap()));
 
 		// Run reducers first
-		ConnectableObservable<RadixNodeAction> reducedNodeActions = nodeActions.doOnNext(action -> {
+		final ConnectableObservable<RadixNodeAction> reducedNodeActions = nodeActions.doOnNext(action -> {
```
mpsandiford: Can the name of this local variable be changed?  It matches the field name at line 64 here, which is a bit awkward.

I'm not really sure why it's not directly assigned to the field in the first place, given that this is the constructor?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -78,6 +86,9 @@ private RadixNetworkController(RadixNetwork network, List<RadixNetworkEpic> epic
 			);
 			*/
```
mpsandiford: Old debug code can probably go now.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomPuller.java
```diff
@@ -49,6 +54,100 @@ public RadixAtomPuller(
...
+				if (fetchAtomsObservationAction.getUuid().equals(initialAction.getUuid())) {
+					if (pullNode == null) {
+						pullNode = action.getNode();
+					}
+
+					AtomObservation observation = fetchAtomsObservationAction.getObservation();
+					if (observation.hasAtom()) {
+						observation.getAtom().spunParticles()
+							.map(s -> TransitionedParticle.fromSpunParticle(s, observation.getType()))
+							.forEach(t -> {
```
mpsandiford: `t` is already a `TransitionedParticle` because of the previous `.map(...)`

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixParticleStore.java
```diff
@@ -1,34 +1,35 @@
...
-				} catch (Throwable e) {
-					emitter.onError(e);
+			.flatMap(observation -> Observable.create(emitter -> {
+				if (observation.hasAtom()) {
+					observation.getAtom().spunParticles().filter(s -> s.getParticle().getShardables().contains(address))
+						.forEach(s -> {
+							final TransitionedParticle tp =
+								TransitionedParticle.fromSpunParticle(s, observation.getType());
+
+							final ParticleObservation p = ParticleObservation.ofParticle(tp);
```
flotothemoon: Petty - technically this should be called `ParticleObservation.ofTransitionedParticle`

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -78,6 +86,9 @@ private RadixNetworkController(RadixNetwork network, List<RadixNetworkEpic> epic
...
 
+			// TODO: Move this into a proper reducer framework
+			reducers.forEach(r -> r.accept(action));
+
 			// TODO: also add equals check
```
flotothemoon: Appreciate this is old - as I understand, this might cause unnecessary updates by calling networkState.onNext when nextState.equals(curState) but nextState != curState (i.e. equals but not same instance). Is this documented anywhere / do we have n issue? Is it a quick fix (looks like it)?

radixdlt-java/src/test/java/com/radixdlt/client/core/ledger/RadixAtomPullerTest.java
```diff
@@ -12,29 +26,134 @@
...
+	public void when_empty_address_pulled__fetch_atoms_action_should_be_emitted_and_head_stored() {
+		RadixNetworkController controller = mock(RadixNetworkController.class);
+		BiConsumer<RadixAddress, AtomObservation> atomStore = mock(BiConsumer.class);
+		RadixAtomPuller radixAtomPuller = new RadixAtomPuller(controller, atomStore);
+		RadixAddress address = mock(RadixAddress.class);
+		RadixNode node = mock(RadixNode.class);
+
+		final AtomicReference<Consumer<RadixNodeAction>> reducer = new AtomicReference<>();
+
+		doAnswer(new Answer() {
```
flotothemoon: lambda here as well

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -94,15 +105,30 @@ private RadixNetworkController(RadixNetwork network, List<RadixNetworkEpic> epic
...
+	 * Actions are only emitted after they have been processed by all reducers.
+	 *
+	 * @return observable of actions in the system
+	 */
 	public Observable<RadixNodeAction> getActions() {
-		return nodeActions;
+		return reducedNodeActions;
 	}
 
 	public void dispatch(RadixNodeAction action) {
```
flotothemoon: Appreciate this is old, should add documentation regardless

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/ParticleObservation.java
```diff
@@ -0,0 +1,32 @@
...
+
+	public TransitionedParticle getParticle() {
+		return particle;
+	}
+
+	public boolean isHead() {
+		return this.particle == null;
+	}
+
+	public static ParticleObservation head() {
```
flotothemoon: Could be a constant i.e.
`public static final ParticleObservation HEAD = new ParticleObservation(null)`

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/ParticleStore.java
```diff
@@ -4,5 +4,5 @@
 import io.reactivex.Observable;
 
 public interface ParticleStore {
-	Observable<TransitionedParticle> getParticles(RadixAddress address);
+	Observable<ParticleObservation> getParticles(RadixAddress address);
```
flotothemoon: Document me

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomObservation.java
```diff
@@ -19,36 +20,60 @@ public static Type fromAtomEventType(AtomEventType type) {
 		}
 	}
 
+	/**
+	 * Describes the type of observation including whether the update is "soft", or a weakly
+	 * supported atom which could possibly be deleted soon
+	 */
+	public static class AtomObservationUpdateType {
```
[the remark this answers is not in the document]
talekhinezh: good point

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
talekhinezh: lol oops

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixParticleStore.java
```diff
@@ -1,34 +1,35 @@
...
-			.flatMap(observation -> Observable.<TransitionedParticle>create(emitter -> {
-				try {
-					observation.getAtom().spunParticles()
-						.filter(s -> s.getParticle().getShardables().contains(address))
-						.forEach(s -> emitter.onNext(TransitionedParticle.fromSpunParticle(s, observation.getType())));
-				} catch (Throwable e) {
-					emitter.onError(e);
+			.flatMap(observation -> Observable.create(emitter -> {
+				if (observation.hasAtom()) {
+					observation.getAtom().spunParticles().filter(s -> s.getParticle().getShardables().contains(address))
```
flotothemoon: Why is it `Atom.addresses()` but `Particle.getShardables`? See https://github.com/radixdlt/radixdlt-java/pull/157/files#diff-fd4469b32bb7ecfbfb76df4695efd053R132

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
talekhinezh: Because the I only want to expose the `connect` interface in the constructor (aka difference between `ConnectableObservable` in the constructor and `Observable` in the class. Any good ideas in the best way of implementing this in cleanest way?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomPuller.java
talekhinezh: That's what I would expect but for some reason compiler is not being happy on my end when trying to run, it thinks `t` is an `Object`. Any ideas why?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomPuller.java
```diff
@@ -49,6 +54,100 @@ public RadixAtomPuller(
...
+			FetchAtomsAction initialAction,
+			ObservableEmitter<AtomObservation> emitter
+		) {
+			this.atomStore = atomStore;
+			this.initialAction = initialAction;
+			this.address = initialAction.getAddress();
+			this.emitter = emitter;
+		}
+
+		// TODO: Replace the following checks with constraint machine to
```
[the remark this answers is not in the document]
talekhinezh: Hard to say until we get more knowledge on how conflicts + atom dependencies will look like.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomPuller.java
```diff
@@ -49,6 +54,100 @@ public RadixAtomPuller(
...
+		@Override
+		public void accept(RadixNodeAction action) {
+			final List<AtomObservation> observations = new ArrayList<>();
+
+			if (action instanceof FetchAtomsObservationAction) {
+
+				FetchAtomsObservationAction fetchAtomsObservationAction = (FetchAtomsObservationAction) action;
+				if (fetchAtomsObservationAction.getUuid().equals(initialAction.getUuid())) {
+					if (pullNode == null) {
+						pullNode = action.getNode();
```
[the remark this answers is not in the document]
talekhinezh: Good catch, was from deprecated code

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixParticleStore.java
talekhinezh: `Atom.addresses` should really be `shardables` but I don't think it's a big deal at the moment

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
talekhinezh: Not sure if it's actually correct to check for `equals` here because epics might be dependent on states being emitted if the reducer processes an action even if the two states are equal. Would have to look into our epics and also what the standard is for this. Either way, this is probably a separate task which I'll add to the backlog.

radixdlt-java/src/test/java/com/radixdlt/client/core/ledger/RadixAtomPullerTest.java
```diff
@@ -12,29 +26,134 @@
...
+	public void when_address_pulled_and_stored_action_emitted__a_soft_store_observation_should_be_stored() {
+		RadixNetworkController controller = mock(RadixNetworkController.class);
+		BiConsumer<RadixAddress, AtomObservation> atomStore = mock(BiConsumer.class);
+		RadixAtomPuller radixAtomPuller = new RadixAtomPuller(controller, atomStore);
+		RadixAddress address = mock(RadixAddress.class);
+		RadixNode node = mock(RadixNode.class);
+
+		final AtomicReference<Consumer<RadixNodeAction>> reducer = new AtomicReference<>();
+
+		doAnswer(new Answer() {
```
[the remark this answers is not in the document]
talekhinezh: Right, there were just some issues in the returned empty lambda but looks like it is solvable with a cast.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
mpsandiford: Ah, OK.  If that's the case, then maybe just a rename of the local variable will be fine. Maybe `connectableReducedNodeActions`?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixAtomPuller.java
mpsandiford: Looks like a type inference fail in the Oracle JDK.  Interestingly the eclipse compiler gets this right.  Best leave it as is I guess.

PR #158, opened 2019-03-13
Add RESYNC into event stream

talekhinezh:
Used so that old cached HEADs aren't used on new pulls

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
@@ -393,7 +393,7 @@ public class RadixApplicationAPI {
 		return this.getActionStore(actionClass).getActions(address, identity)
 			.map(actionClass::cast)
 			.publish()
-			.autoConnect()
+			.refCount()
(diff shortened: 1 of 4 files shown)


radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -393,7 +393,7 @@ public TokenTypeReference getNativeTokenRef() {
 		return this.getActionStore(actionClass).getActions(address, identity)
 			.map(actionClass::cast)
 			.publish()
-			.autoConnect()
+			.refCount()
```
laszlourszuly: Just a self-educational question: the difference here is that `refCount()` also disconnects the underlying observable when the last observer is detached, whereas the `autoConnect()` only guarantees that the underlying observable is connected (but not disconnected), right?

talekhinezh: Yes, exactly. And after last observer is detached and then a new observer subscribes, `refCount` will start a new flow while `autoConnect` will still be connected to the original.

PR #159, opened 2019-03-14
Expose ActionExecutionException errorData

laszlourszuly:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ActionExecutionException.java
```diff
@@ -13,16 +13,30 @@
...
-		super(reasons.toString());
+		super(reasons != null && !reasons.isEmpty()
+                ? reasons.toString()
+                : String.valueOf(errorData));
 
 		this.errorData = errorData;
 		this.reasons = Collections.unmodifiableList(reasons);
 	}
 
+    /**
```
mpsandiford: Indentation here and a few other places seems to be spaces rather than tabs.

mpsandiford:
No issues with this, and I think it's something that should be available.  Apart from the indentation tidy up, which isn't a dealbreaker for me, I don't think there is any changes I would want to see here.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ActionExecutionException.java
```diff
@@ -13,16 +13,30 @@
 	private final JsonObject errorData;
 
 	private ActionExecutionException(JsonObject errorData, List<ActionExecutionExceptionReason> reasons) {
-		super(reasons.toString());
+		super(reasons != null && !reasons.isEmpty()
+                ? reasons.toString()
+                : String.valueOf(errorData));
 
 		this.errorData = errorData;
 		this.reasons = Collections.unmodifiableList(reasons);
```
mpsandiford: I know this is old code, but this looks wrong to me.  It's not possible to make a List immutable post-hoc by doing this kind of thing, as the original passed-in list is still mutable and mutating it will mutate our (supposedly) unmodifiable list.

Better would be `this.reasons = ImmutableList.copyOf(reasons);`

laszlourszuly: Would it be the Guava `com.google.common.collect.ImmutableList`?

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ActionExecutionException.java
```diff
@@ -13,16 +13,30 @@
 	private final JsonObject errorData;
 
 	private ActionExecutionException(JsonObject errorData, List<ActionExecutionExceptionReason> reasons) {
-		super(reasons.toString());
+		super(reasons != null && !reasons.isEmpty()
+		? reasons.toString()
```
flotothemoon: Shouldn't this be intended?

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ActionExecutionException.java
```diff
@@ -13,16 +13,30 @@
...
 	public List<ActionExecutionExceptionReason> getReasons() {
 		return reasons;
 	}
 
+	/**
+	 * @return The optional JSON representation of the raw error data. May be empty but never null.
+	 */
+	public JsonObject getErrorData() {
+		return errorData != null
+		? errorData.deepCopy()
```
flotothemoon: same, shouldn't this be intended?

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ActionExecutionException.java
laszlourszuly: Ah! These spaces vs. tabs are killing me :)

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ActionExecutionException.java
mpsandiford: Yes, that is the one, otherwise it's
```
this.reasons = Collections.unmodifiableList(new ArrayList<>(reasons));
````

PR #162, opened 2019-03-20
Feature/use big decimal

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/CreateTokenToParticleGroupsMapper.java
```diff
@@ -45,27 +46,27 @@
...
+			ImmutableMap.of(
+				MintedTokensParticle.class, mintPermissions,
+				BurnedTokensParticle.class, burnPermissions,
+				TransferredTokensParticle.class, TokenPermission.ALL
+			),
+			null
 		);
 
-		if (tokenCreation.getInitialSupply().isZero()) {
+		if (tokenCreation.getInitialSupply().compareTo(BigDecimal.ZERO) <= 0) {
```
mpsandiford: I think a negative initial supply should be an error, not silently truncated to zero.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/CreateTokenAction.java
```diff
@@ -16,23 +15,23 @@
...
-		UInt256 initialSupply,
-		UInt256 granularity,
+		BigDecimal initialSupply,
+		BigDecimal granularity,
 		TokenSupplyType tokenSupplyType
 	) {
 		// Redundant null check added for completeness
 		Objects.requireNonNull(initialSupply);
-		if (tokenSupplyType.equals(TokenSupplyType.FIXED) && initialSupply.isZero()) {
+		if (tokenSupplyType.equals(TokenSupplyType.FIXED) && initialSupply.compareTo(BigDecimal.ZERO) <= 0) {
```
mpsandiford: initialSupply < 0 should be an error, rather than silently accepted here.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenUnitConvert.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import org.radix.utils.UInt256;
+import org.radix.utils.UInt256s;
+
+/**
+ * Utility class for converting token units between UInt256 and BigDecimal
+ */
+public final class TokenUnitConvert {
```
mpsandiford: Should probably be called `TokenUnitConversions`.  Would then be called as `TokenUnitConversions.subunitsToUnits(foo)` which reads better IMO.

Otherwise `ConvertTokenUnits` would be fine too, but has the disadvantage that it doesn't start with `Token`.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/CreateTokenToParticleGroupsMapper.java
talekhinezh: yikes, thanks for catching that

PR #164, opened 2019-03-21
Deteministic Signatures using Low value S

Sajjon:
Changing ECDSA signatures to be deterministic according to [RFC6979](https://tools.ietf.org/html/rfc6979) in order to more easily write tests and compare implementations between Radix libraries. Also ensuring that signatures are using [Low value S (BIP-62)](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures).

Both JS and Swift library use deterministic Signatures and Low S. Now we can compare Signatures across libraries which is very advantageous.


Sajjon:
Closed in favor of split PR,
See 
#166 
and
#167

PR #165, opened 2019-03-21
Feature/application shard state

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ShardedAppStateId.java
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (!(o instanceof ShardedAppStateId)) {
+			return false;
+		}
+
+		ShardedAppStateId r = (ShardedAppStateId) o;
+		return r.stateClass.equals(stateClass) && r.address.equals(address);
```
mpsandiford: Would be better to use `Objects.equals(a, b)` as this is a bit more resistant to NPE's.  Appreciate that fields are final and non-null per constructor, but this can help make things more robust under change.

PR #166, opened 2019-03-21
Low S in signatures according to BIP62

Sajjon:
Adding boolean flag for signing method to ensure Low value S in signatures [according to BIP62](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#Low_S_values_in_signatures)


mpsandiford:
I think we should merge this, as it brings the library up to the same functional standard as RadixCore.  Any other thoughts on this @talekhinezh , @flotothemoon ?

PR #169, opened 2019-03-22
Feature/remove static universe

talekhinezh:


examples/src/main/java/com/radixdlt/client/examples/ChatBot.java
```diff
@@ -63,12 +63,10 @@ public Completable apply(DecryptedMessage message) {
 	}
 
 	public static void main(String[] args) throws Exception {
-		RadixUniverse.bootstrap(Bootstrap.BETANET);
```
flotothemoon: Yes! YES! Finally! :D

examples/src/main/java/com/radixdlt/client/examples/RadixWalletExample.java
```diff
@@ -33,7 +27,7 @@ public static void main(String[] args) throws Exception {
 			radixIdentity = RadixIdentities.loadOrCreateFile("my.key");
 		}
 
-		RadixApplicationAPI api = RadixApplicationAPI.create(radixIdentity);
+		RadixApplicationAPI api = RadixApplicationAPI.create(Bootstrap.LOCALHOST, radixIdentity);
```
flotothemoon: Some inconsistent formatting with the other example

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/Ownable.java
```diff
@@ -1,10 +1,12 @@
...
 
+import com.radixdlt.client.atommodel.accounts.RadixAddress;
 import com.radixdlt.client.core.crypto.ECPublicKey;
 
 /**
  * Temporary interface for representing a particle that can be owned and only invalidated by an owner.
  */
 public interface Ownable {
 	ECPublicKey getOwner();
+	RadixAddress getAddress();
```
flotothemoon: Hmm I think it's weird we have both, we shouldn't need both - can we remove getAddress() and just make `getOwner` return `RadixAddress`

radixdlt-java/src/main/java/com/radixdlt/client/core/Bootstrap.java
```diff
@@ -2,19 +2,28 @@
...
 public enum Bootstrap implements BootstrapConfig {
-	BETANET(
+	LOCALHOST(
 		RadixUniverseConfigs::getBetanet,
 		Observable.just(
 			new RadixNode("localhost", false, 8080),
 			new RadixNode("localhost", false, 8081)
 		)
 	),
+	LOCALHOST_SINGLENODE(
```
flotothemoon: Why are these different?

talekhinezh: SINGLE_NODE will always connect to the same node so tests which have two clients can be synced easily

PR #172, opened 2019-03-25
Bugfix/rlau 1071 token symbol conflicts

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenDefinitionsReducer.java
```diff
@@ -58,17 +57,30 @@ public TokenDefinitionsState reduce(TokenDefinitionsState state, TransitionedPar
...
+						: mintedOrBurnedAmount
+				);
 
-			TokenDefinitionReference tokenDefinitionReference = p instanceof MintedTokensParticle
-				? ((MintedTokensParticle) p).getTokenDefinitionReference() : ((BurnedTokensParticle) p).getTokenDefinitionReference();
-			return state.mergeSupplyChange(tokenDefinitionReference, change);
+				TokenDefinitionReference tokenDefinitionReference = p instanceof MintedTokensParticle
+					? ((MintedTokensParticle) p).getTokenDefinitionReference() : ((BurnedTokensParticle) p).getTokenDefinitionReference();
+				return state.mergeSupplyChange(tokenDefinitionReference, change);
+			} else if (t.getTransition() == ParticleTransition.U2N) {
```
mpsandiford: It seems to me that these two are really the same thing, but with different signs.  In an ideal world, they would be the same code.  For me this isn't a deal breaker, but maybe could be tidied up.

flotothemoon: This logic could be simplified a bit by merging the two branches and negating the logic?

flotothemoon: Ah yes, you were faster, ha. Deleting my comment..

talekhinezh: I won't worry about cleaning this up as I think experience here shows that this rewindability of state should probably be handled at a lower layer.

PR #173, opened 2019-03-26
Feature/rlau 1070 remove feeparticle

flotothemoon:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/FeeMapper.java
```diff
@@ -4,9 +4,12 @@
...
 import com.radixdlt.client.core.crypto.ECPublicKey;
+import org.radix.common.tuples.Pair;
 
 import java.util.List;
+import java.util.Map;
 
 public interface FeeMapper {
-	List<ParticleGroup> map(Atom atom, RadixUniverse universe, ECPublicKey key);
+	// TODO maybe all Mappers should be able to return metadata as well..?
+	Pair<Map<String, String>, List<ParticleGroup>> map(Atom atom, RadixUniverse universe, ECPublicKey key);
```
mpsandiford: Would prefer to see this as a more specific interface rather than an extended aggregation of generic types with arguments.  Something like AtomFees with a getMetaDataEntrys() and getParticleGroupEntries() would be more illuminating here.

flotothemoon: Yes, I was thinking that as well but I don't think we know the abstractions well enough yet, so I just stuck with a generic type. Don't want to get stuck with a specific type too soon.

PR #175, opened 2019-03-28
Bugfix/rlau 1083 inter dependency consensus

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/CreateTokenToParticleGroupsMapper.java
```diff
@@ -66,7 +66,7 @@
 				TokenUnitConversions.unitsToSubunits(tokenCreation.getInitialSupply()),
 				TokenUnitConversions.unitsToSubunits(tokenCreation.getGranularity()),
 				tokenCreation.getAddress(),
-				System.currentTimeMillis(),
+				System.nanoTime(),
```
flotothemoon: Why nano time?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/RadixParticleStore.java
```diff
@@ -18,7 +23,14 @@ public RadixParticleStore(AtomStore atomStore) {
 		return cache.computeIfAbsent(address, addr -> atomStore.getAtoms(address)
 			.flatMap(observation -> Observable.create(emitter -> {
 				if (observation.hasAtom()) {
-					observation.getAtom().spunParticles().filter(s -> s.getParticle().getShardables().contains(address))
+					final List<SpunParticle> spunParticleList = observation.getAtom().spunParticles().collect(Collectors.toList());
+
+					if (observation.getType() == Type.DELETE) {
```
flotothemoon: Is this only because Core doesn't return DELETE Atoms in the right order? 
If so, we should make a task for that

talekhinezh: No, it's because currently particle reducers can only handle particle DELETEs if done in reverse order due to the way it manages state.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/CreateTokenToParticleGroupsMapper.java
talekhinezh: Found that using currentTimeMillis could sometimes return back the same if called in succession close to each other.

mpsandiford: Also, currentTimeMillis() is often quantised as well.  On some OS's only updated in units of 18ms or so.

PR #176, opened 2019-03-28
Fix for DiscoverSingleNodeEpic in case of universe mismatch

mpsandiford:
Plus additional debug logging and updated universe.


mpsandiford:
Looks good to merge now.

PR #182, opened 2019-04-11
Add ECSignature constructor which takes bytes as an argument

MarcRubio:


radixdlt-java/src/main/java/com/radixdlt/client/core/crypto/ECSignature.java
```diff
@@ -25,6 +29,21 @@ public ECSignature(BigInteger r, BigInteger s) {
 		this.s = Bytes.trimLeadingZeros(s.toByteArray());
 	}
 
+	public ECSignature(byte[] bytes) {
+		DLSequence seq;
+		ASN1Integer r, s;
+		try (ASN1InputStream decoder = new ASN1InputStream(bytes)) {
```
flotothemoon: Don't really have a problem with *how* this works, but would like to see this in a more explicit function instead. So something like ECSignature.fromASN1(bytes) or whatever it's called, which then internally calls to ECSignature(r, s)

laszlourszuly:
@MarcRubio This change has been reviewed and approved. Do you mind if we merge it?

Sajjon:
@MarcRubio I'm late to the party, but the `fromANS1` method uses DER encoding which is the broader term. ANS1 is just a standard interface, the but the actual encoding is most likely DER. So I think it should be renamed. Or if it isn't DER then it should be stated if it is ZER or PER. More [info on Wiki](https://en.wikipedia.org/wiki/Abstract_Syntax_Notation_One)

MarcRubio:
Yes you are right. Dan originally had this method as decodeFromDER so  it should be fromDER or decodeFromDER. I went for fromANS1 due to Florians suggestion. This may be refactored out of this class but I should be changed nevertheless.

PR #186, opened 2019-04-17
Feature/handle deletes

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ParticleReducer.java
```diff
@@ -1,14 +1,19 @@
...
-import com.radixdlt.client.core.ledger.TransitionedParticle;
+import com.radixdlt.client.core.atoms.particles.Particle;
 
 /**
- * Java version of redux-like reducer where particles are the actions
+ * Temporary class for reducing up particles into an application state.
+ * TODO: Introduce the AtomOS layer in between Application and Particles
+ * TODO: which will reduce fungible, indexable, and payload particle types
+ * TODO: into a state machine upon which application state can be built on
+ * TODO: top of
```
laszlourszuly: God bless for this comment!
I love when a developer cares enough for his/her audience to give a few lines of context where needed.

talekhinezh:
@laszlourszuly Yes, the reason for moving from async to sync at the dapp action layer is that the creation of state machine transitions from a given state should really be synchronous logic, i.e. given this state give me a transition. Makes it much cleaner/easier to test. Any async stuff should really be done at a different layer.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -185,12 +188,9 @@ private RadixApplicationAPI(
...
 		this.actionStores = atomMappers.stream().collect(Collectors.toMap(
 			AtomToExecutedActionsMapper::actionClass,
-			m -> new ActionStore<>(ledger.getAtomStore(), m)
-		));
-		this.applicationStores = particleReducers.stream().collect(Collectors.toMap(
-			ParticleReducer::stateClass,
-			r -> new ApplicationStore<>(ledger.getParticleStore(), r)
+			m -> m
 		));
+		this.applicationStores = particleReducers.stream().collect(Collectors.toMap(ParticleReducer::stateClass, r -> r));
```
flotothemoon: Can there only be one reducer per state class?

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/ParticleReducer.java
flotothemoon: +1
Please let's not call it 'payload' as discussed

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenDefinitionsState.java
```diff
@@ -17,32 +17,48 @@
...
+		HashMap<RRI, TokenState> newState = new HashMap<>(state);
 		final TokenState tokenState = state.get(ref);
-		Map<EUID, UnallocatedTokensParticle> newTokenUnallocated = new HashMap<>(tokenState.getUnallocatedTokens());
-		if (add) {
-			newTokenUnallocated.put(unallocatedTokensParticle.getHid(), unallocatedTokensParticle);
-		} else {
-			newTokenUnallocated.remove(unallocatedTokensParticle.getHid());
+		if (tokenState == null) {
+			newState.put(ref, new TokenState(
+				null,
```
flotothemoon: null null null? 
Should add a comment

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenState.java
```diff
@@ -44,6 +44,28 @@ public TokenState(
...
+				.putAll(state1.unallocatedTokens)
+				.build();
+
+		UInt256 unallocatedAmount = unallocated.entrySet().stream()
+			.map(Entry::getValue)
+			.map(UnallocatedTokensParticle::getAmount)
+			.reduce(UInt256.ZERO, UInt256::add);
+
+		return new TokenState(
+			state0.name != null ? state0.name : state1.name,
```
flotothemoon: So, we state0 takes precedence over state1? Why?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -1,39 +1,56 @@
...
+						return;
+					}
 
-	/**
-	 * The In Memory Atom Data Store
-	 */
-	private final ConcurrentHashMap<RadixAddress, Subject<AtomObservation>> cache = new ConcurrentHashMap<>();
+					if (observation.getUpdateType().getType() == Type.STORE || !observation.getUpdateType().isSoft()) {
+						// This first so that leaves get deleted first
+						softDeleteDependentsOf(observation.getAtom());
```
flotothemoon: Is this threadsafe? I realise you're using concurrent maps here, but I'm wondering if there is a case where we could end up in an invalid state because gathering and deletion is done in the same step.

On another note, what is the point of using concurrent maps (with the overhead they introduce) when all methods accessing them are `synchronised` anyway?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -43,102 +60,131 @@
...
-
-						// Should always update observation state if going from soft to hard observation
-						final boolean isSoftToHard = lastUpdate != null && lastUpdate.isSoft() && !nextUpdate.isSoft();
-
-						if (include || isSoftToHard) {
-							atomsObservationState.put(observation);
-							missingDeletes.forEach(atomsObservationState::put);
+							return Stream.empty();
 						}
+					}).reduce(Spin.NEUTRAL, (s0, s1) -> SpinStateMachine.isAfter(s0, s1) ? s0 : s1) == Spin.UP
```
flotothemoon: Could make some performance optimisations here given the fact that we only need to know that it has not been DOWNed, not really critical though

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -43,102 +60,131 @@
...
-						final boolean include = atomsObservationState.curCount >= observationCountPerAddress.getOrDefault(address, 0L);
-						return include ? Observable.just(observation) : Observable.empty();
-					} else {
-						return Observable.empty();
+	@Override
+	public Observable<AtomObservation> getAtomObservations(RadixAddress address) {
+		return Observable.create(emitter -> {
+			synchronized (lock) {
+				final CopyOnWriteArrayList<ObservableEmitter<AtomObservation>> observers;
+				if (!allObservers.containsKey(address)) {
```
flotothemoon: Nitpicking, would prefer to see this as 
```
observers = allObservers.get(address);
if (observers == null) {
 // create and put
}
```

flotothemoon: (saves an unnecessary lookup)

radixdlt-java/src/main/java/com/radixdlt/client/dapps/messaging/RadixMessaging.java
```diff
@@ -1,50 +0,0 @@
...
-import com.radixdlt.client.application.RadixApplicationAPI.Result;
-import com.radixdlt.client.atommodel.accounts.RadixAddress;
-
-import io.reactivex.Observable;
-import io.reactivex.observables.GroupedObservable;
-
-/**
- * High Level API for Instant Messaging. Currently being used by the Radix Android Mobile Wallet.
- */
-public class RadixMessaging {
```
flotothemoon: Why was this removed? Are there plans to reintroduce this in some other form?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: yes

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenState.java
talekhinezh: it doesn't matter which one because intrinsically they should be the same. This will get cleaned up once we insert the layer in between application state and particles.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
talekhinezh: Yes, good point, no need for concurrent maps now. But we shouldn't be getting into invalid state because deletes are just flagged rather than actually being removed from the map.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
talekhinezh: Performance isn't concern right now but correctness. This class is going to change a lot to get the right modules in the right place.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
talekhinezh: but then you lose `final`

radixdlt-java/src/main/java/com/radixdlt/client/dapps/messaging/RadixMessaging.java
talekhinezh: all functionality here is available via `RadixApplicationAPI`

PR #189, opened 2019-04-19
Fix/androidjava8methodscomp

MarcRubio:
Small changes to reduce API compatibility in Android from API 26 (Oreo) to  API 24 (Nougat)


radixdlt-java/src/main/java/org/radix/utils/primitives/Longs.java
```diff
@@ -61,7 +61,7 @@ public static long fromByteArray(byte[] bytes) {
 	 * @param bytes The byte array to decode to a long
 	 * @param offset The offset within the array to start decoding
 	 * @return The decoded long value
-	 */
+	 * */
```
mpsandiford: Minor comment format fail.

MarcRubio: How did that even happen??? Anyways, fixed.

PR #192, opened 2019-05-20
Feature/staging groups

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -833,21 +832,53 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
...
-			.lastOrError()
-			.map(this::buildAtomWithFee);
+	/**
+	 * Returns a cold single of an unsigned atom given a user action. Note that this is
+	 * method will always return a unique atom even if given equivalent actions
+	 *
+	 * @param actions actions to build a single atom
+	 * @return a cold single of an atom mapped from an action
+	 */
+	public Single<UnsignedAtom> buildAtom(List<Action> actions) {
```
mpsandiford: Would be more generally useful as `Iterable<Action>` rather than `List<Action>`.

Iterable can be converted to stream easily: `StreamSupport.stream(actions.spliterator(), false)`.

Will also need to change `stageActions(...)` method too.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -825,6 +768,62 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
...
+		public void execute(Action action) {
+			this.actions.add(action);
+		}
+
+		/**
+		 * Commit the transaction onto the ledger
+		 * @return the results of committing
+		 */
+		public Result commit() {
+			return buildDisconnectedResult(actions.toArray(new Action[0])).connect();
```
mpsandiford: Would be better to have `buildDisconnectedResult(Action...)` take an `Iterable`, as this particular use case takes an Iterable, converts it to an array, only to have buildDisconnectedResult turn it back into an Iterable again.

There is another use in `execute(Action)` below that would need to be wrapped in a singleton, but currently it's wrapped in an array reference anyway, so I don't think this is a strong negative.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -825,6 +768,62 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
...
+		}
+	}
+
+	/**
+	 * Create a new transaction
+	 *
+	 * @return a new transaction
+	 */
+	public Transaction transaction() {
```
mpsandiford: Maybe should be called newTransaction or createTransaction.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -896,8 +927,8 @@ private Result buildDisconnectedAtomSubmit(Atom atom) {
 		return buildDisconnectedAtomSubmit(Single.just(atom));
 	}
 
-	private Result buildDisconnectedResult(Action action) {
-		final Single<Atom> atom = this.buildAtom(action)
+	private Result buildDisconnectedResult(Action... actions) {
```
mpsandiford: Would prefer to see this taking a `Iterable<Action>` as mentioned elsewhere.

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenBalanceState.java
```diff
@@ -6,85 +6,25 @@
...
 	}
 
-	public TokenBalanceState(Map<RRI, Balance> balance) {
+	public TokenBalanceState(Map<RRI, BigDecimal> balance) {
 		this.balance = ImmutableMap.copyOf(balance);
 	}
 
-	public Map<RRI, Balance> getBalance() {
+	public Map<RRI, BigDecimal> getBalance() {
 		return Collections.unmodifiableMap(balance);
```
mpsandiford: Not sure the ImmutableMap needs to be wrapped as an unmodifiableMap :)

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
```diff
@@ -48,4 +43,23 @@
 	 * @return a never ending observable of updates
 	 */
 	Observable<AtomObservation> getAtomObservations(RadixAddress address);
+
+	/**
+	 * Retrieve the current set of validated up particles at a given shardable.
+	 * @param address the address to get the particles under
+	 * @return a stream of all up particles of the current local view
```
mpsandiford: Missing UUID documentation, so I don't know what the intent of this parameter is.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
```diff
@@ -48,4 +43,23 @@
...
+	 */
+	Stream<Particle> getUpParticles(RadixAddress address, String uuid);
+
+	/**
+	 * Adds the particle group to the staging area
+	 * @param particleGroup the particle group to add to staging area
+	 */
+	void stageParticleGroup(String uuid, ParticleGroup particleGroup);
```
mpsandiford: Missing uuid doc.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
```diff
@@ -48,4 +43,23 @@
...
+
+	/**
+	 * Retrieves all staged particle groups and clears
+	 * @return all staged particle groups in order
+	 */
+	List<ParticleGroup> getStagedAndClear(String uuid);
```
mpsandiford: Missing uuid doc.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
```diff
@@ -48,4 +43,23 @@
...
+
+	/**
+	 * Adds the particle group to the staging area
+	 * @param particleGroup the particle group to add to staging area
+	 */
+	void stageParticleGroup(String uuid, ParticleGroup particleGroup);
+
+	/**
+	 * Retrieves all staged particle groups and clears
+	 * @return all staged particle groups in order
```
mpsandiford: What order?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -825,6 +768,62 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
 		return new UnsignedAtom(new Atom(allParticleGroups, metaData));
 	}
 
+	/**
+	 * Represents an atomic transaction to be committed to the ledger
+	 */
+	public final class Transaction {
```
flotothemoon: So.. an atom??

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
flotothemoon: Shouldn't this be static? 
And agreed, would rather have it be just new()

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
laszlourszuly: This one can't be static because of dependency on `buildDisconnectedResult(Action[])`, right?

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -833,21 +832,53 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
...
 
-		return Observable.concat(statelessParticleGroups, statefulParticleGroups.flatMapIterable(l -> l))
-			.<List<ParticleGroup>>scanWith(
-					ArrayList::new,
-					(a, b) -> Stream.concat(a.stream(), Stream.of(b)).collect(Collectors.toList())
-			)
-			.lastOrError()
-			.map(this::buildAtomWithFee);
+	/**
+	 * Returns a cold single of an unsigned atom given a user action. Note that this is
```
flotothemoon: given user action*s*

radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/TokenBalanceState.java
flotothemoon: It's "extra" safe. Like using two condoms..

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -179,9 +214,9 @@ public void store(RadixAddress address, AtomObservation atomObservation) {
 
 
 	@Override
-	public Stream<Particle> getUpParticles(RadixAddress address) {
+	public Stream<Particle> getUpParticles(RadixAddress address, String stagedUuid) {
 		synchronized (lock) {
-			return particleIndex.entrySet().stream()
+			Set<Particle> upParticles = particleIndex.entrySet().stream()
```
flotothemoon: It seems potentially dangerous that this does not preserve ordering - what's the reasoning behind this?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -193,10 +228,25 @@ public void store(RadixAddress address, AtomObservation atomObservation) {
...
+
 					Set<Atom> uppingAtoms = spinParticleIndex.getOrDefault(Spin.UP, Collections.emptySet());
 					return uppingAtoms.stream().anyMatch(a -> atoms.get(a).isStore());
 				})
-				.map(Map.Entry::getKey);
+				.map(Map.Entry::getKey)
+				.collect(Collectors.toSet());
+
+			if (stagedUuid != null) {
+				stagedParticleIndex.getOrDefault(stagedUuid, Collections.emptyMap()).entrySet().stream()
```
flotothemoon: In this case could just use .ifPresent instead of .getOrDefault

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
laszlourszuly: Yup, I would agree with this comment.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
flotothemoon: Documenting this is probably enough - the returned Stream would indicate *to me* that it's ordered in some meaningful way

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -32,6 +36,9 @@
 	private final Object lock = new Object();
 	private final Map<RadixAddress, Boolean> syncedMap = new HashMap<>();
 
+	private final Map<String, Atom> stagedAtoms = new ConcurrentHashMap<>();
```
flotothemoon: Why are we using atoms here? Just as a container for particle groups + timestamp?

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
```diff
@@ -48,4 +43,23 @@
 	 * @return a never ending observable of updates
 	 */
 	Observable<AtomObservation> getAtomObservations(RadixAddress address);
+
+	/**
+	 * Retrieve the current set of validated up particles at a given shardable.
+	 * @param address the address to get the particles under
+	 * @return a stream of all up particles of the current local view
+	 */
+	Stream<Particle> getUpParticles(RadixAddress address, String uuid);
```
flotothemoon: I think we should add a default variant with uuid as null
Also could add `Nullable` annotation to uuid

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -53,6 +60,34 @@ private void softDeleteDependentsOf(Atom atom) {
...
+	public void stageParticleGroup(String uuid, ParticleGroup particleGroup) {
+		synchronized (lock) {
+			Atom stagedAtom = stagedAtoms.get(uuid);
+			if (stagedAtom == null) {
+				stagedAtom = new Atom(particleGroup, System.currentTimeMillis());
+			} else {
+				List<ParticleGroup> groups = Stream.concat(stagedAtom.particleGroups(), Stream.of(particleGroup)).collect(Collectors.toList());
+				stagedAtom = new Atom(groups, System.currentTimeMillis());
+			}
+			stagedAtoms.put(uuid, stagedAtom);
```
flotothemoon: danger: uuid could be null, null keys are not permitted

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
flotothemoon: order of insertion I would presume, but please clarify

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -825,6 +768,62 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
...
+	/**
+	 * Create a new transaction
+	 *
+	 * @return a new transaction
+	 */
+	public Transaction transaction() {
+		return new Transaction();
+	}
+
+	private void stageActions(String uuid, List<Action> actions) {
```
flotothemoon: Can this be null here
`AtomStore.stageParticleGroup` should not accept null values, so this probably shouldn't either

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
flotothemoon: I fear saying "Transaction" might be confusing to people, since database transaction != financial transaction and it's often confused in DLTs

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
```diff
@@ -53,6 +60,34 @@ private void softDeleteDependentsOf(Atom atom) {
...
+				stagedParticleIndex.put(uuid, index);
+			});
+		}
+	}
+
+	@Override
+	public List<ParticleGroup> getStagedAndClear(String uuid) {
+		synchronized (lock) {
+			final Atom atom = stagedAtoms.remove(uuid);
+			stagedParticleIndex.get(uuid).clear();
```
flotothemoon: danger: uuid could be null and concurrent hash map does not permit null keys, so this will NPE

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -825,6 +768,62 @@ public UnsignedAtom buildAtomWithFee(List<ParticleGroup> particleGroups) {
...
+	public final class Transaction {
+		private final ArrayList<Action> actions = new ArrayList<>();
+		private Transaction() {
+		}
+
+		/**
+		 * Execute an action within this transaction
+		 * @param action the action to execute
+		 */
+		public void execute(Action action) {
```
laszlourszuly: I have just reviewed the regression tests where I got slightly confused by the wording "execute". I first thought it really meant "execute" (as in "do something - now"). I then remembered that "execute" in this case means "add" or "enqueue" or "stash" (no, not stash) or "save-for-later". Maybe we could rename this method to something that better reflects what's actually going on. `addAction` could be a classical alternative.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
laszlourszuly: I associate this `Transaction` with a financial transaction and I think it works for me and the use-case I had in mind. Then again; maybe I'm just a special kind of stupid who got lucky with a "mental double negation" this time...

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
laszlourszuly: Also could this return `this` to allow a builder (-ish) pattern?

```java
public Completable prepareTransaction() {
    return api.transaction()
            .execute(new SomeAction())
            .execute(new SomeOtherAction())
            .commit()
            .toCompletable();
}
```

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/InMemoryAtomStore.java
talekhinezh: Not sure what you mean, streams don't imply ordering they can even be parallel

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
talekhinezh: Actually not so sure here now. `actions` should really be an ordered list, using `Iterable`would encourage Sets and Collections to be used which really shouldn't be done.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
laszlourszuly: Fair point!

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -836,14 +837,14 @@ private void stageActions(String uuid, List<Action> actions) {
 	}
 
 	/**
-	 * Returns a cold single of an unsigned atom given a user action. Note that this is
-	 * method will always return a unique atom even if given equivalent actions
+	 * Returns a cold single of an unsigned atom given an ordered iterable of user actions.
+	 * Note that this is method will always return a unique atom even if given equivalent actions
```
mpsandiford: Too much "is" in this sentence I think.

radixdlt-java/src/main/java/com/radixdlt/client/core/ledger/AtomStore.java
```diff
@@ -6,6 +6,7 @@
 import com.radixdlt.client.core.atoms.ParticleGroup;
 import com.radixdlt.client.core.atoms.particles.Particle;
 import io.reactivex.Observable;
+import io.reactivex.annotations.Nullable;
```
mpsandiford: Would prefer the javax.annotation.Nullable version.

PR #198, opened 2019-06-05
Remove TokenDefinitionParticle icon field, replacing with an icon_url field

mpsandiford:


radixdlt-java/src/main/java/com/radixdlt/client/atommodel/tokens/TokenDefinitionParticle.java
```diff
@@ -43,9 +43,9 @@
 	@DsonOutput(Output.ALL)
 	private UInt256 granularity;
 
-	@JsonProperty("icon")
+	@JsonProperty("icon_url")
```
Sajjon: We are mixing snake and camelCase a lot, but mostly we do camelCase (I think?), can we maybe decide to stick to always be camelCasing our JSON Keys? :) 

**If** there is a **complete tie** in opinions between camelCase and snake_case, we should go for camelCase since it saves A LOT of code in the Swift world, talking about a couple of hundreds of lines of code, maybe more...

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/tokens/TokenDefinitionParticle.java
```diff
@@ -69,7 +69,7 @@ public TokenDefinitionParticle(
 		this.description = description;
 		this.granularity = granularity;
 		this.tokenPermissions = ImmutableMap.copyOf(tokenPermissions);
-		this.icon = icon;
+		this.iconUrl = iconUrl;
```
Sajjon: should we perform any validation at all? Should we allow the empty string? null? a MASSIVE string? Should we validate that it is indeed a valid URL?

radixdlt-java/src/main/resources/universe/betanet.json
```diff
@@ -48,6 +48,7 @@
                         "spin": 1,
                         "serializer": "radix.spun_particle",
                         "particle": {
+                            "icon_url": ":str:https://www.radixdlt.com/wp-content/uploads/2019/04/radix-logo.svg",
```
flotothemoon: should check with marketing if that's really the logo we want to use here
maybe it should be a permanent link like radixdlt.com/icon where there's always the most recent version

mpsandiford: That for sure isn't the icon that we want to use going forward.  I have an issue in the backlog to update this once we have a specific icon and location we want to use.

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/tokens/TokenDefinitionParticle.java
mpsandiford: The RadixCore software ensures that, if present, the URL is valid.  I don't think it makes much sense to duplicate that validation here, as if the application is accepting arbitrary input from users, it will likely want to do some validation there to ensure that inputs are acceptable.

Null is permitted.  Feel free to use as large a string as you can afford :)

radixdlt-java/src/main/java/com/radixdlt/client/atommodel/tokens/TokenDefinitionParticle.java
mpsandiford: I'll change it to camelCase.  That seems to be the general consensus anyway from other comments.

PR #206, opened 2019-06-20
New betanet-staging and sunstone universes

blenessy:
Stay off sunstone for a while - doing stability tests there (you will not like it).


blenessy:
@x-wf When a universe is change I usually create a PR to make it easy for the dev team to integrate.

PR #208, opened 2019-06-24
Bugfix/fix unsupported exception

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -120,8 +120,14 @@ private RadixNetworkController(
 			.map(epic -> epic.epic(connectableReducedNodeActions, networkState))
 			.collect(Collectors.toSet());
 
-		// FIXME: Cleanup disposable
-		Observable.merge(updates).subscribe(this::dispatch);
+		// FIXME: Cleanup disposable and error handling
```
laszlourszuly: Is this `FIXME` still valid?

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
```diff
@@ -120,8 +120,14 @@ private RadixNetworkController(
...
+		Observable.merge(updates).subscribe(
+			this::dispatch,
+			e -> {
+				e.printStackTrace();
```
laszlourszuly: Do we have a dedicated logger (like `slf4j` or similar) that could/would/should handle the printing of this exception instead? (Benefit being that the logger can be configured "globally" to control output to some degree, while `printStackTrace()` doesn't honour any of those configurations).

radixdlt-java/src/main/java/com/radixdlt/client/core/network/actions/DiscoverMoreNodesErrorAction.java
```diff
@@ -0,0 +1,30 @@
...
+
+	public Throwable getReason() {
+		return reason;
+	}
+
+	// TODO: remove this
+	@Override
+	public RadixNode getNode() {
+		throw new UnsupportedOperationException();
+	}
```
laszlourszuly: Oh, the ruthless humour of the Universe 😂

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/DiscoverNodesEpic.java
```diff
@@ -34,9 +35,9 @@ public DiscoverNodesEpic(Observable<RadixNode> seeds, RadixUniverseConfig config
 	public Observable<RadixNodeAction> epic(Observable<RadixNodeAction> updates, Observable<RadixNetworkState> networkState) {
 		Observable<RadixNodeAction> getSeedUniverses = updates
 			.ofType(DiscoverMoreNodesAction.class)
-			.firstOrError()
-			.flatMapObservable(i -> seeds)
-			.map(GetUniverseRequestAction::of);
+			.flatMap(i -> seeds)
+			.<RadixNodeAction>map(GetUniverseRequestAction::of)
+			.onErrorReturn(DiscoverMoreNodesErrorAction::new);
```
laszlourszuly: Is this `onErrorReturn(...)` for not breaking the observable stream (as a hard `onError` would) and thereby saving me the effort of having to "reconnect" after each error?

Beautiful, if so!

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
talekhinezh: Yes, should really not do a `subscribe` inside constructor but it has been good enough for now.

radixdlt-java/src/main/java/com/radixdlt/client/core/network/RadixNetworkController.java
talekhinezh: good point

radixdlt-java/src/main/java/com/radixdlt/client/core/network/epics/DiscoverNodesEpic.java
talekhinezh: Yup, exactly. It's RX version of try/catch.

PR #209, opened 2019-06-25
Add types for sonar warning and simplify code.

mpsandiford:


mpsandiford:
> Yup, LGTM (except for the failing tests, that is :) )

Failing tests were issues with SonarQube failing to do its job for some reason.

PR #211, opened 2019-06-26
Feature/staging areas

talekhinezh:


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/StageActionException.java
```diff
@@ -0,0 +1,10 @@
+package com.radixdlt.client.application.translate;
+
+/**
+ * An exception which occurs during the construction of an action.
```
flotothemoon: construction or staging of an action?

PR #219, opened 2019-07-17
Fix decoding of attachment in a transaction

MarcRubio:
Attachment was not being decoded from a Base64 String

### DIFF
--- a/radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/AtomToTokenTransfersMapper.java
+++ b/radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/AtomToTokenTransfersMapper.java
@@ -1,25 +1,23 @@
 package com.radixdlt.client.application.translate.tokens;
 
+import com.google.gson.JsonParser;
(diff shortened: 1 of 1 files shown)


radixdlt-java/src/main/java/com/radixdlt/client/application/translate/tokens/AtomToTokenTransfersMapper.java
```diff
@@ -1,26 +1,24 @@
 package com.radixdlt.client.application.translate.tokens;
 
+import com.google.gson.JsonParser;
```
mpsandiford: Would prefer if we didn't unnecessarily change the order of the import statements.  Just messes up diffs in the future.

MarcRubio: woops sorry, I simply did the auto imports that intelliJ does ctrl + option + O to remove the unused RadixConstants import

I'll let you guys decide on this ordering thing for the lib but in Android I try and do this auto imports shortcut on all files for consistency

PR #222, opened 2019-07-23
Adds ability to retrieve icon URL from TokenState

mpsandiford:


mpsandiford:
> Looks great! I even agree with the ordering of the method arguments :)
> 
> One question though: The corresponding changes to the public API (`api.createToken(...)`, `CreateTokenAction.create(...)` et al) will come in a separate change, right?

Yes, that's covered in a separate issue.

PR #235, opened 2019-10-02
Feature/add jenkins six nodes config

salvatop:
added config to run six nodes network on jenkins


salvatop:
@mpsandiford can we merge this?

clstrfsck:
> @mpsandiford can we merge this?

Can you talk to @talekhinezh about how to do this?  The code looks fine, but the target for the merge is wrong.  It should be either rc/1.0-beta.5 (which doesn't exist yet), or we should rename the source for the merge to hotfix/[something-or-other].

PR #240, opened 2019-10-21
Tempo 2 Code Cleanup

pgyschuk:


flotothemoon:
This PR removes classes that (while unused) are unrelated to Tempo 2.0. Further, some of the unused classes are very useful for other purposes (e.g. TestUtils), even though they are not directly referenced in production code.

PR #243, opened 2019-10-31
BouncyCastle initialization refactoring

pgyschuk:


radixdlt-java/src/main/java/com/radixdlt/client/application/identity/PrivateKeyEncrypter.java
```diff
@@ -64,7 +65,7 @@ public static String createEncryptedPrivateKey(String password) throws GeneralSe
 
         SecretKey derivedKey = getSecretKey(password, salt, ITERATIONS, KEY_LENGTH);
 
-        Cipher cipher = Cipher.getInstance("AES/CTR/NoPadding");
+        Cipher cipher = Cipher.getInstance("AES/CTR/NoPadding", "BC");
```
pgyschuk: If provider is not specified it will be selected from list of available providers. On Radix.java we have configuration of BouncyCastle so looks like it is preferable provider. But if for some reason this provider will not be available and we will not specify that it must be used system will use any provider which supporting specified algorithm. To make it more clear I decided to add "BC" parameter for all invocations where this provider have to be used.

radixdlt-java/src/test/java/com/radixdlt/client/application/identity/RadixIdentitiesTest.java
```diff
@@ -12,7 +12,7 @@
 	@Test
 	public void newEncryptedIdentityWriterTest() throws Exception {
 		Writer writer = mock(Writer.class);
-		RadixIdentities.createNewEncryptedIdentity(writer, "");
+		RadixIdentities.createNewEncryptedIdentity(writer, "Password");
```
pgyschuk, line 15: Interesting case - BouncyCastle do not allow empty password for key generation it failing with Empty Password exception. But if security provider is defined explicitly it will fail without fallback. I believe it is better to know which provider is used for key generation

radixdlt-java/src/test/java/com/radixdlt/client/core/address/RadixAddressTest.java
```diff
@@ -22,7 +23,7 @@ public void createAddressFromPublicKey() {
 		assertEquals(address, RadixAddress.from("JHB89drvftPj6zVCNjnaijURk8D8AMFw4mVja19aoBGmRXWchnJ"));
 	}
 
-	@Test(expected = IllegalArgumentException.class)
```
pgyschuk, line 25: new version of library throwing `DecoderException` instead of `IllegalArgumentException` in case of invalid key

radixdlt-java/src/main/java/com/radixdlt/client/application/identity/PrivateKeyEncrypter.java
clstrfsck: As this is a library, I don't think we should be forcing folks to use the "BC" provider, which is what this does.

Would suggest that we use the default provider, and allow consuming applications to set up their provider list as they would like.

clstrfsck (changes requested):
Overall looks OK, but I don't think that we should be forcing applications to use specific security providers.

radixdlt-java/src/main/java/com/radixdlt/client/application/identity/PrivateKeyEncrypter.java
pgyschuk: fixed

mpsandiford:
Looks good to me.

clstrfsck (approved):
Looks good to me, thanks!

PR #244, opened 2019-11-20
Refactoring of Atom model

pgyschuk:


radixdlt-java/src/main/resources/universe/localnet.json
```diff
@@ -1,9 +1,9 @@
 {
   "magic": -1332248574,
-  "hid": ":uid:71b120ee41cd3d7d93be8b2267244595",
```
pgyschuk, line 3: In this PR Universe is regenerated because temporalProof were removed

PR #285, opened 2020-09-30
Switch to factory methods

siy:


radixdlt-java/build.gradle
```diff
@@ -68,8 +68,30 @@ jacocoTestReport {
     }
 }
 
+configurations.all {
+    resolutionStrategy.eachDependency {
```
clstrfsck, line 72: Looks like a copy/paste duplicate?

clstrfsck (approved):
Looks good to me, assuming the duplicated `configurations.all` section is removed from radixdlt-java/build.gradle.

Approving on the basis that you are going to fix this.

radixdlt-java/build.gradle
siy: Thanks for spotting it. Fixed.

PR #296, opened 2020-11-02
Add api to calculate atom minimum required fee

LukasGasior1:


radixdlt-java/src/main/java/com/radixdlt/fees/FeeTable.java
```diff
@@ -50,6 +55,15 @@ public UInt256 minimumFee() {
 		return this.feeEntries;
 	}
 
+	public UInt256 feeFor(Atom atom) {
+		final int feeSize = Serialize.getInstance().toDson(atom, DsonOutput.Output.HASH).length;
```
LukasGasior1: I'm not entirely sure about moving this to FeeTable, another option I've considered is to expose this method in FeeProcessor

siy: Name for `feeSize` does not look good to me. At first look I've considered it somehow related to fee itself rather that part of fee related to atom size. Perhaps something like 'atomSize' would be more clear and will better reflect the value.

siy (approved):
LGTM with comments

radixdlt-java/src/main/java/com/radixdlt/fees/FeeTable.java
```diff
@@ -50,6 +55,15 @@ public UInt256 minimumFee() {
 		return this.feeEntries;
 	}
 
+	public UInt256 feeFor(Atom atom) {
```
clstrfsck: Might be worth considering changing the name of the argument to `atomWithoutFees` to make it clear that this will not correctly calculate the fee for an atom that already has fees included.

radixdlt-java/src/main/java/com/radixdlt/fees/FeeTable.java
```diff
@@ -50,6 +55,15 @@ public UInt256 minimumFee() {
...
+		final int feeSize = Serialize.getInstance().toDson(atom, DsonOutput.Output.HASH).length;
+		final ImmutableSet<Particle> outputs = atom.spunParticles()
+				.filter(sp -> Spin.UP.equals(sp.getSpin()))
+				.map(SpunParticle::getParticle)
+				.collect(ImmutableSet.toImmutableSet());
+		return feeFor(atom, outputs, feeSize);
+	}
+
 	public UInt256 feeFor(Atom atom, Set<Particle> outputs, int feeSize) {
```
clstrfsck: Similarly here (understood this is not new code).

clstrfsck (approved):
Looks good.  Some suggestions not affecting functionality.

radixdlt-java/src/main/java/com/radixdlt/client/application/RadixApplicationAPI.java
```diff
@@ -1099,6 +1099,10 @@ public void discoverNodes() {
 		return this.universe.getNetworkController().getActions();
 	}
 
+	public BigDecimal getMinimumRequiredFee(Atom atom) {
```
clstrfsck: Also here should be "atomWithoutFees".  Sorry missed this one the first time round.

clstrfsck (approved):
Just a comment on the API method which requires an atom without fees included to function properly.  All good to merge once this is addressed.

siy (commented):
LGTM

PR #298, opened 2020-11-17
Add SystemParticle to java lib

LukasGasior1:
SystemParticle.java is copied over from engine-library


siy (approved):
LGTM

clstrfsck (approved):
Would like to see at least an EqualsVerifier test for this, but otherwise looks good.

